PR #13, opened 2020-03-09
Datahub integration fetch metadata

saruniitr:
## Change description

This PR adds changes to enable Data Hub integration in the tool.
It add `ACCESS_TOKEN`, `HAWK` authentication related keys to the environment which enables the tool to use Data Hub services and fetch metadata.

Helper functions are added to do Company search, contact search from the tool. These are useful at the time of Investment enquiry creation to see if a company/contact already exists.

Other functions added to fetch uuids for various investment related fields. The reference data in the tool are mapped to corresponding uuid before they are sent to Data Hub.

These changes are mainly in preparation for the another PR in which the actual investment enquiry will be posted. This is split out to keep the size of PR shorter, if everything is preferred in a single PR please let me know, I will update this accordingly.


app/enquiries/common/datahub_utils.py
```diff
@@ -0,0 +1,182 @@
...
+    Each request has a timeout (default=15sec) failing which throws an
+    exception which will be captured in Sentry
+    """
+
+    if request_headers:
+        headers = request_headers
+    else:
+        headers = {
+            "Content-Type": "application/json",
+            "Authorization": f"Bearer {settings.DATA_HUB_ACCESS_TOKEN}",
```
reupen: The access token for the OAuth endpoints comes from Staff SSO – it isn't an application one, it identifies a user.

saruniitr: Yes agree, SSO integration hasn't happened yet so for me to test I am sending it in the headers.

app/enquiries/common/datahub_utils.py
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+    try:
+        response = requests.request(
+            method, url, headers=headers, data=payload, timeout=timeout
```
reupen: Would suggest passing `json=<data structure>` rather than `data=json.dumps(...)`.

saruniitr: ok, agree it is better

app/enquiries/tests/test_dh_integration.py
```diff
@@ -0,0 +1,23 @@
...
+class DataHubIntegrationTests(TestCase):
+    @mock.patch("requests.request")
+    def test_dh_request_timeout(self, mock_request):
+        """ Tests to ensure data hub requests raise exception """
+        mock_request.side_effect = Timeout
+        url = settings.DATA_HUB_COMPANY_SEARCH_URL
+        payload = {"name": "test"}
+
+        with pytest.raises(Timeout):
+            response = dh_request("POST", url, payload, timeout=2)
```
reupen: Would consider using `requests-mock` rather than patching `requests` directly – it will give more realistic behaviour: https://requests-mock.readthedocs.io/en/latest/

app/enquiries/tasks.py
```diff
@@ -0,0 +1,31 @@
...
+
+    with Redis(host=redis_host) as rds:
+        try:
+            rds.ping()
+
+            dh_metadata = dh_fetch_metadata()
+            rds.set(key, json.dumps(dh_metadata))
+            logging.info(f"Data Hub metadata last refreshed at {datetime.now()}")
+        except ConnectionError:
+            logging.error("Error connecting to Redis backend, cannot fetch metadata")
```
reupen: I would consider using the Django cache framework rather than using Redis directly: 

- https://docs.djangoproject.com/en/3.0/topics/cache/
- http://jazzband.github.io/django-redis/latest/

app/enquiries/common/datahub_utils.py
```diff
@@ -0,0 +1,182 @@
...
+import os
+import requests
+
+from django.conf import settings
+from mohawk import Sender
+from requests.exceptions import RequestException
+from rest_framework import status
+
+
+def dh_request(method, url, payload, request_headers=None, timeout=15):
```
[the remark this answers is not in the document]
saruniitr: This is a preparatory PR adding mostly helper functions accessing Data hub API. The next PR uses these functions to create investments, my plan is to include more tests in that PR.

Also I really do not want to add tests that test Data hub API because they already exist hence added a test that simulates failure.
Besides unit testing these functions any suggestions on what tests you would like to see?

app/enquiries/tasks.py
```diff
@@ -0,0 +1,31 @@
...
+    ignore_result=True,
+)
+def refresh_datahub_metadata(redis_host="redis", key="metadata"):
+
+    with Redis(host=redis_host) as rds:
+        try:
+            rds.ping()
+
+            dh_metadata = dh_fetch_metadata()
+            rds.set(key, json.dumps(dh_metadata))
```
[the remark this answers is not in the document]
saruniitr: Yes that is the plan. I was not keen to couple caching changes in that function but since it needs to cached in some way it makes sense to handle caching within that function. I will modify this.

app/enquiries/tasks.py
saruniitr: definitely simplifies, modified.

app/enquiries/tasks.py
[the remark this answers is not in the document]
saruniitr: updated function to handle caching also.

app/enquiries/common/datahub_utils.py
```diff
@@ -0,0 +1,155 @@
...
+        "sector",
+    )
+
+    credentials = settings.HAWK_CREDENTIALS[settings.HAWK_ID]
+
+    metadata = {"failed": []}
+    for endpoint in endpoints:
+        meta_url = os.path.join(settings.DATA_HUB_METADATA_URL, endpoint)
+
+        print(f"Fetching {meta_url} ...")
```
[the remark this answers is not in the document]
saruniitr: updated with logging module functions

app/enquiries/tests/test_dh_integration.py
saruniitr: Couldn't get it to work, will revisit later but for now will keep the current approach.

saruniitr:
Addressed review comments,
- Used django cache backend instead of using redis directly
- moved cache handling from periodic task to the fetch function
- basic unit tests as more tests is tricky in this case as explained in above comment. Appreciate any suggestions to improve this.

app/settings.py
```diff
@@ -173,7 +173,46 @@
...
+# Celery and Redis
+DATA_HUB_METADATA_FETCH_INTERVAL_HOURS=env.int('DATA_HUB_METADATA_FETCH_INTERVAL_HOURS', default=4)
+REDIS_BASE_URL = env('REDIS_BASE_URL', default=None)
+if REDIS_BASE_URL:
+    REDIS_CELERY_DB = env('REDIS_CELERY_DB', default=1)
+    BROKER_URL = f'{REDIS_BASE_URL}/{REDIS_CELERY_DB}'
+    CELERY_RESULT_BACKEND = BROKER_URL
+    CELERY_ACCEPT_CONTENT = ['application/json']
+    CELERY_TASK_SERIALIZER = 'json'
+    CELERY_RESULT_SERIALIZER = 'json'
```
reupen: These are all the default values, there should be no reason to override them:
docs.celeryproject.org/en/latest/userguide/configuration.html

app/settings.py
```diff
@@ -173,7 +173,46 @@
...
+dh_metadata = {}
+DATA_HUB_ACCESS_TOKEN=env('DATA_HUB_ACCESS_TOKEN')
+DATA_HUB_METADATA_URL=env('DATA_HUB_METADATA_URL')
+DATA_HUB_COMPANY_SEARCH_URL=env('DATA_HUB_COMPANY_SEARCH_URL')
+DATA_HUB_CONTACT_SEARCH_URL=env('DATA_HUB_CONTACT_SEARCH_URL')
+
+HAWK_ID=env("HAWK_ID")
+HAWK_CREDENTIALS = {
+    env("HAWK_ID"): {
+        "id": env("HAWK_ID"),
```
reupen: Is storing `HAWK_ID` in three places needed?

app/enquiries/tasks.py
```diff
@@ -0,0 +1,32 @@
...
+
+    try:
+        # set expiry few minutes before next refresh so that we
+        # ensure refresh fetch data again
+        expiry_secs = settings.DATA_HUB_METADATA_FETCH_INTERVAL_HOURS * 60 * 60 - (5 * 60)
+        dh_metadata = dh_fetch_metadata(expiry_secs=expiry_secs)
+        logging.info(f"Data Hub metadata last refreshed at {datetime.now()}")
+    except Exception as e:
+        logging.error(f"Error refreshing metadata, {e}")
+        raise e
```
reupen: I'm not sure why `logging.error()` is needed – if the task fails, Celery will log the error, and we'll end up will double-Sentry alerts.

app/enquiries/common/datahub_utils.py
```diff
@@ -0,0 +1,204 @@
...
+    """
+
+    dh_data = list(
+        filter(lambda d: d[target_key] == refdata_value, dh_metadata[dh_category])
+    )
+
+    return dh_data[0]["id"] if dh_data else None
+
+
+def dh_company_search(company_name):
```
reupen: Is this meant to be in this PR?

app/enquiries/common/datahub_utils.py
```diff
@@ -0,0 +1,204 @@
...
+                    "postcode": address["postcode"],
+                    "country": address["country"]["name"],
+                },
+            }
+        )
+
+    return companies
+
+
+def dh_contact_search(contact_name):
```
reupen: Is this meant to be in this PR?

app/settings.py
saruniitr: these are included for completeness, they can be removed.

app/settings.py
saruniitr: not really, I was trying group them an object but only ID and KEY are sufficient, will modify this.

app/enquiries/tasks.py
saruniitr: My understanding is sentry only monitors exceptions? yes but if errors also then I can remove this.

app/enquiries/common/datahub_utils.py
reupen: If that's the case, please add a TODO here and in the configuration file, so it doesn't cause any confusion later on (for other people).

app/enquiries/common/datahub_utils.py
saruniitr: Yes, these are the main function in this PR. We will perform a company search and provide user with the matching results (assuming there are matches in DH). User can select the appropriate one and assign it to the enquiry. This gives an early indication/confidence to the user that enquiry is for a valid company and they can create a company if not found.

app/enquiries/common/datahub_utils.py
saruniitr: same as above

app/enquiries/common/datahub_utils.py
saruniitr: ok

app/enquiries/tasks.py
reupen: The logging integration is enabled by default in Sentry: https://docs.sentry.io/platforms/python/logging/

However, the Django and Celery integrations will probably need turning on (if they haven't been already).

saruniitr:
Please let me know if there are any other comments

app/enquiries/tasks.py
saruniitr: removed error log

app/enquiries/common/datahub_utils.py
```diff
@@ -0,0 +1,214 @@
...
+            content_type=None,
+            always_hash_content=False,
+        )
+        response = requests.get(
+            meta_url, headers={"Authorization": sender.request_header}, timeout=10
+        )
+        if response.ok:
+            metadata[endpoint] = response.json()
+        else:
+            metadata["failed"].append(endpoint)
```
reupen: This will end up being cached – so what will happen when the page is displayed if some of the metadata requests failed?

app/enquiries/common/datahub_utils.py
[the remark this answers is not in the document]
reupen: There is quite a lot of logic in here about caching, error handling and transforming results. I really think some tests are warranted.

reupen:
> Yes, besides fetching metadata I wanted to include few other helper functions. I could've split but many other branches are dependent on this PR getting merged, it will further push review/merge of other branches.

If this is blocking other things I would suggest splitting things out. For example, company and contact search don't need to be part of this PR. As for metadata, I think the first step would be to add metadata retrieval without caching and add the caching afterwards.

app/enquiries/celery.py
```diff
@@ -0,0 +1,12 @@
...
+from django.conf import settings
+
+# set the default Django settings module for the 'celery' program.
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'app.settings')
+app = Celery('app')
+
+# Using a string here means the worker will not have to
+# pickle the object when using Windows.
+app.config_from_object('django.conf:settings')
+app.autodiscover_tasks(lambda: settings.INSTALLED_APPS)
```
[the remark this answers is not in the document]
saruniitr: well it is not really required here but we can provide a list of packages to `autodiscover_tasks`
https://docs.celeryproject.org/en/stable/reference/celery.html

app/enquiries/common/datahub_utils.py
```diff
@@ -0,0 +1,214 @@
...
+
+    It check if the data is valid in cache, if it has expired then fetches again
+    """
+
+    try:
+        cached_metadata = cache.get(cache_key)
+        if not cached_metadata:
+            logging.info("Metadata expired in cache, fetching again ...")
+            cached_metadata = _dh_fetch_metadata()
+            cache.set(cache_key, json.dumps(cached_metadata), timeout=expiry_secs)
```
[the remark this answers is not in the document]
saruniitr: yes, but using the data structure is better here, modified it.

app/enquiries/tasks.py
```diff
@@ -0,0 +1,31 @@
...
+def refresh_datahub_metadata():
+
+    try:
+        # set expiry few minutes before next refresh so that we
+        # ensure refresh fetch data again
+        expiry_secs = settings.DATA_HUB_METADATA_FETCH_INTERVAL_HOURS * 60 * 60 - (5 * 60)
+        dh_metadata = dh_fetch_metadata(expiry_secs=expiry_secs)
+        logging.info(f"Data Hub metadata last refreshed at {datetime.now()}")
+    except Exception as e:
+        raise e
```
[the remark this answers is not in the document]
saruniitr: yes, now it can be removed

app/settings.py
```diff
@@ -173,7 +173,37 @@
 
 STATIC_URL = '/static/'
 
-
```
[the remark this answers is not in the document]
saruniitr: enabled integrations

app/enquiries/common/datahub_utils.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+    companies = []
+    url = settings.DATA_HUB_COMPANY_SEARCH_URL
+    payload = {"name": company_name}
+
+    response = dh_request("POST", url, payload)
+
+    # It is not an error for us if the request fails, this can happen if the
+    # Access token is invalid, consider that there are no matches
+    if response.status_code != status.HTTP_200_OK:
```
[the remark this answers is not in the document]
saruniitr: done

app/enquiries/common/datahub_utils.py
```diff
@@ -0,0 +1,214 @@
...
+    Returns list of subset of fields for each company found
+    """
+    companies = []
+    url = settings.DATA_HUB_COMPANY_SEARCH_URL
+    payload = {"name": company_name}
+
+    response = dh_request("POST", url, payload)
+
+    # It is not an error for us if the request fails, this can happen if the
+    # Access token is invalid, consider that there are no matches
```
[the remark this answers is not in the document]
saruniitr: This is to specifically handle the case if access token is not valid then we get authentication credentials not provided error. This is an error wrt Data Hub api but not a fatal one for the application. We can assume it is same not finding any matches but the error information is propagated upwards (which gets shown to user). Perhaps the comment is not clear.
Company search is done when user edits enquiry to see if there are any matches. If there is a problem at this point we won't get results but user is notified of the error and can take action.

This won't happen once SSO is integrated as user will be automatically redirected to login if token expires, since it is not merged yet I am handling this here.

app/enquiries/common/datahub_utils.py
saruniitr: Yes but it is safe as it is only intended for debug purposes, no decisions are being taken based on this field. The code that actually uses these changes is not using this field. Once these changes are merged I can actually use these functions and see if it can be removed. For now I will retain this.

app/enquiries/common/datahub_utils.py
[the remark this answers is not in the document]
saruniitr: I have added tests to cover caching and connection issues. I am going add few more cases to cover company and contact search, hopefully that is sufficient for this PR.
As I mentioned earlier, this is preparation work and the next PR actually uses these functions. I will extend these tests in that PR as that gives complete picture.

saruniitr:
I think it is better not to split at this stage because all of these functions are required for next PR. Only fetching metadata or company/contact search is not sufficient. Splitting now pushes things again. I think the tests situation has improved and will extend them further in the next PR.

app/enquiries/tasks.py
```diff
@@ -0,0 +1,30 @@
...
+from redis.exceptions import ConnectionError
+
+from app.enquiries.common.datahub_utils import dh_fetch_metadata
+
+FETCH_INTERVAL_HOURS = f"*/{settings.DATA_HUB_METADATA_FETCH_INTERVAL_HOURS}"
+
+
+@periodic_task(
+    run_every=(crontab(minute="*/1")),
+    # run_every=(crontab(hour=FETCH_INTERVAL_HOURS, minute="0")),
```
reupen: Could you check this – wasn't the commented out line correct?

saruniitr: yes commented line is correct, it was done for debug but forgot to remove while pushing these changes. I will correct this.

saruniitr:
thanks for the approval and I agree with your comment on metadata caching, I will be adding more tests in that area in next PR where investment creation feature uses this metadata.

saruniitr:
Thanks for the reviews, received two approvals, merging it.

PR #15, opened 2020-03-10
Add Service healthcheck route

saruniitr:
## Change description

This PR adds a route that reports the Service status.
Pingdom is going to ping this route to get service status.

Service is healthy (OK) if we can connect to the database otherwise reported as ERROR.
If we can connect to the database then we check if there are any enquiries exists and provide a comment in the response accordingly.

Also includes a unit test to test the route.


saruniitr:
Received two approvals, thanks for your reviews, merging it.

PR #17, opened 2020-03-11
Feature/rtt 66 manual CSV import/upload backend

kiddhustle:
This PR adds the server side functionality for manual CSV imports of enquiries detail in this Jira ticket:

https://uktrade.atlassian.net/browse/RTT-66

It also adds basic views to aid with testing (i.e. success page and error redirects).

There is a separate ticket for the full UI of the CSV imports which will be submitted as a separate PR. Details can be found here: https://uktrade.atlassian.net/browse/RTT-55

So PLEASE IGNORE the HTML templates


app/enquiries/tests/test_views.py
```diff
@@ -260,3 +269,110 @@ def test_enquiry_detail_template_ref_data(self):
...
+
+        response = self.client.post(
+            reverse("import-enquiries"), {"enquiries": upload}, follow=True
+        )
+
+        final_count = Enquiry.objects.count()
+        self.assertNotContains(response, ImportEnquiriesView.ERROR_HEADER)
+        self.assertContains(response, f"Successfully posted {num_items} records!")
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(final_count, num_items)
```
elcct: Would be good to check if the objects were imported correctly, not just how many we ended up with.

app/enquiries/utils.py
```diff
@@ -0,0 +1,40 @@
+import app.enquiries.ref_data as ref_data
+from app.enquiries.models import Enquirer, Enquiry
+
+
+def row_to_enquiry(row: dict) -> Enquirer:
```
elcct: I think there is no test for this function.

app/enquiries/views.py
```diff
@@ -58,3 +74,61 @@ def form_invalid(self, form):
...
+                        records.append(row_to_enquiry(row))
+        logger.info(f"Successfully ingested {len(records)} records")
+        return records
+    def post(self, request, *args, **kwargs):
+        error_url = reverse("import-enquiries") + '?errors=1'
+        records = []
+        enquiries_key = "enquiries"
+
+        try:
+            if enquiries_key in request.FILES or enquiries_key in request.POST:
```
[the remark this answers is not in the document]
elcct: Can you explain more about test environment? When I run upload tests, the file appears in the `request.FILES` and also when I upload the file from the browser.

kiddhustle:
@elcct You're right about the request.POST the file is now showing request.FILES across all envs (I must have fixed the problem I was having along the way)

Have also added more test re you feedback today

app/enquiries/views.py
```diff
@@ -99,3 +116,66 @@ def form_invalid(self, form):
...
+            logger.error(err)
+            return HttpResponseRedirect(self.ERROR_URL)
+        return render(
+            self.request,
+            "import-enquiries-confirmation.html",
+            {"enquiries": records, "ERROR_HEADER": self.ERROR_HEADER},
+        )
+
+    def get(self, request, *args, **kwargs):
+        status_code = 400 if "errors" in request.GET else 200
```
currycoder: This seems a little weird to me.  If I'm reading it correctly, this view returns a 400 if the calling request tells it to e.g. `GET /enquiries/import/?errors=true`

Can you help me to understand it?

app/enquiries/views.py
```diff
@@ -99,3 +116,66 @@ def form_invalid(self, form):
...
+                messages.error(
+                    self.request,
+                    f"File is not of type: text/csv with  extension .csv. Detected type: {f.content_type}",
+                )
+                return HttpResponseRedirect(reverse("import-enquiries"))
+            with transaction.atomic():
+                for c in f.chunks(chunk_size=settings.UPLOAD_CHUNK_SIZE):
+                    csv_file = csv.DictReader(codecs.iterdecode(BytesIO(c), "utf-8"))
+                    for row in csv_file:
+                        records.append(row_to_enquiry(row))
```
currycoder: This is fairly heavily indented which makes it a little difficult to follow what's going on.  Would it be possible to firstly read all of the chunks for the file in to a buffer and then separately create the csv reader and iterate over it.  e.g. something like the following pseudo code?

```
with uploaded file as f:
     buffered_file = self.._buffer_upload(f)

csv_file = csv.DictReader(buffered_file)
with transaction.atomic():
    for row in csv_file:
        records.append(row_to_enquiry(row))
```

app/enquiries/tests/test_views.py
```diff
@@ -260,3 +270,118 @@ def test_enquiry_detail_template_ref_data(self):
...
+        final_count = db_enquiries.count()
+        self.assertNotContains(response, ImportEnquiriesView.ERROR_HEADER)
+        self.assertContains(response, f"Successfully posted {num_items} records!")
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(final_count, num_items)
+
+        for e in enquiries:
+            # generate filter keyword args dynamically
+            # (can't modify the dict we arfe iterating over withouit raising an error so make a extra copy)
+            qs_kwargs = utils.csv_row_to_enquiry_filter_kwargs(e)
```
elcct: You could probably just do it this way, so there would be no need for a function:
```
csv_row = {
  'enquirer_last_name': 'last_name',
  'phone': '1234567',
}

qs_kwargs = {
  key.replace('enquirer_', 'enquirer__'): value
  for key, value in csv_row.items()
}

print(qs_kwargs)
```

kiddhustle: I didn't think of this.

But then it would still be good to have as a function I am using the the same function in a test for another PR

app/enquiries/management/commands/export_enquiries.py
```diff
@@ -0,0 +1,39 @@
...
+    help = 'this command exports enquiries to simple (level) CSV'
+
+    def handle(self, *args, **options):
+        count = 0
+        entries = Enquiry.objects.all().iterator()
+        with open(OUTPUT_FILE_NAME, 'w', newline='') as f:
+            writer = csv.writer(f)
+            # write headers
+            writer.writerow(ref_data.IMPORT_COL_NAMES)
+            for e in entries:
```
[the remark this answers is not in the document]
kiddhustle: also it has been agreed with @rahulbist that we will export all enquiries

app/enquiries/views.py
kiddhustle: Anything UI related to this functionality of this branch is just for dev/testing purposes.

I just needed a view to make sure that errors are being handled in some way. This ill change once the PR for the UI element is approved (after this one is)

app/enquiries/views.py
kiddhustle: This approach was taken becasue we want to avaoid reading all of the file in one go.

We don't know how big the file is that that user will import so we read it in smaller chucks. This way we can avaoid reaching memory limits.

app/enquiries/tests/test_views.py
```diff
@@ -1,23 +1,31 @@
+import csv
 import pytest
 import pytz
 import random
 
+from io import StringIO, BytesIO
```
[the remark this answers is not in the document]
kiddhustle: removed

app/enquiries/utils.py
```diff
@@ -0,0 +1,58 @@
...
+        job_title=row["enquirer_job_title"],
+        email=row["enquirer_email"],
+        phone=row["enquirer_phone"],
+        request_for_call=row["enquirer_request_for_call"],
+    )
+
+    # validate enquirer before saving - https://docs.djangoproject.com/en/3.0/ref/models/instances/#django.db.models.Model.full_clean
+    enquirer.full_clean()
+
+    e = Enquiry(
```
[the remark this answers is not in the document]
kiddhustle: amended

app/enquiries/tests/test_views.py
kiddhustle: amended in function

app/enquiries/management/commands/export_enquiries.py
```diff
@@ -0,0 +1,39 @@
+import csv
+from django.core.management.base import BaseCommand
+from app.enquiries.models import Enquiry
+import app.enquiries.ref_data as ref_data
+
+OUTPUT_FILE_NAME = 'ingest_sample.csv'
+
+class Command(BaseCommand):
```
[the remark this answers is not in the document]
kiddhustle: amended

app/enquiries/tests/test_views.py
kiddhustle: amended

app/enquiries/utils.py
kiddhustle: amended

app/enquiries/views.py
currycoder: OK that makes sense, but my comment about the level of indentation being difficult to follow here remains.  To help with that, can I suggest breaking out the transaction/chunking block in to a helper method?

e.g.
```python
with uploaded_file as f:
    if not f.name.endswith(".csv") or f.content_type != "text/csv":
        messages.error(
            self.request,
            f"File is not of type: text/csv with  extension .csv. Detected type: {f.content_type}",
        )
        return HttpResponseRedirect(reverse("import-enquiries"))
    records = self._build_enquiries(f)
```

saruniitr:
Thanks for reviews, two approvals received, merging it.

PR #18, opened 2020-03-11
Feature/rtt 73 import template download endpoint

kiddhustle:
This PR adds the Import spreadsheet (.xlsx) template endpoint which allows IST users to download a template where they can manually enter enquiries as a team.

The main business case for this is that a large proportion of enquiries (roughly 50%) come via LinkedIn which users can only export as a CSV file. Also this matches the users existing workflow.

User workflow using the solution:
1. Download the template
2. User enters enquiries into the spreadsheet
3. User exports the enquiries as .CSV
4. User uploads the file into the enquiries tool.

Some of this functionality overlaps with PR-17 so ideally this would be merged first

The ticket can be found here: https://uktrade.atlassian.net/browse/RTT-73

The UI for this feature will be implemented in a separate PR (https://uktrade.atlassian.net/browse/RTT-55)

This includes a management command, to be run by developer if/when they want to generate a template locally to generate a .csv for local development/testing (in my case I created this command to save our BA from having to copy and paste of our valid option from the Git repo manually).

### DIFF
--- a/app/enquiries/ref_data.py
+++ b/app/enquiries/ref_data.py
@@ -70,7 +70,7 @@ class PrimarySector(models.TextChoices):
     DEFAULT = "DEFAULT", _("----")
     ADVANCED_ENG = "ADVANCED_ENG", _("Advanced Engineering")
     AEROSPACE = "AEROSPACE", _("Aerospace")
-    AGRICULTURE = "AGRICULTURE,", _("Agriculture, Horticulture, Fisheries and Pets")
+    AGRICULTURE = "AGRICULTURE", _("Agriculture, Horticulture, Fisheries and Pets")
     AIRPORTS = "AIRPORTS", _("Airports")
     AUTOMOTIVE = "AUTOMOTIVE", _("Automotive")
     CHEMICALS = "CHEMICALS", _("Chemicals")
--- a/app/enquiries/ref_data.py
+++ b/app/enquiries/ref_data.py
@@ -426,6 +426,23 @@ class DatahubProjectStatus(models.TextChoices):
     DELAYED = "DELAYED", _("Delayed")
 
 
+IMPORT_COL_NAMES = [
+    "enquirer_first_name",
+    "enquirer_last_name",
+    "enquirer_job_title",
+    "enquirer_email",
+    "enquirer_phone",
+    "enquirer_request_for_call",
+    "country",
+    "company_name",
+    "primary_sector",
+    "company_hq_address",
+    "website",
+    "investment_readiness",
+    "enquiry_text",
+    "notes",
+]
+
 MAP_ENQUIRY_FIELD_TO_REF_DATA = {
     "enquiry_stage": EnquiryStage,
     "investment_readiness": InvestmentReadiness,
--- a/app/enquiries/tests/test_views.py
+++ b/app/enquiries/tests/test_views.py
@@ -3,12 +3,7 @@ import pytz
 import random
 from openpyxl import Workbook, load_workbook
 
-<<<<<<< HEAD
 from datetime import date, datetime
-=======
-from datetime import date
-
->>>>>>> add spreadsheet download endpoint and tests
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.forms.models import model_to_dict
@@ -18,11 +13,8 @@ from rest_framework import status
 from unittest import mock
 
 import app.enquiries.ref_data as ref_data
-<<<<<<< HEAD
 import app.enquiries.tests.utils as test_utils
-=======
 import app.enquiries.views as enquiry_views
->>>>>>> add spreadsheet download endpoint and tests
 
 from app.enquiries.models import Enquiry, Enquirer
 from app.enquiries.tests.factories import (
--- a/app/enquiries/tests/test_views.py
+++ b/app/enquiries/tests/test_views.py
@@ -1,8 +1,14 @@
 import pytest
 import pytz
 import random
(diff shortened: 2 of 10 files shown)


app/enquiries/tests/test_views.py
```diff
@@ -96,13 +100,20 @@ def test_enquiry_detail_template_simple(self):
...
         country_display_name = get_display_name(ref_data.Country, enquiry.country)
         self.assertContains(response, enquiry_stage_display_name)
-        self.assertContains(response, country_display_name)
\ No newline at end of file
+        self.assertContains(response, country_display_name)
+
+    def test_import_template(self):
+        """Helper function to get a single enquiry and retrieve details"""
+        response = self.client.get(reverse("import-template"))
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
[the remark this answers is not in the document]
currycoder: Can't you check that the file that is responded is of the correct format?  And includes expected reference data etc.

app/enquiries/views.py
```diff
@@ -58,3 +61,16 @@ def form_invalid(self, form):
...
+class ImportTemplateDownloadView(View):
+    methods = ["get"]
+    CONTENT_TYPE = "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
+
+    def get(self, request):
+        response = HttpResponse(content_type=self.CONTENT_TYPE)
+        response[
+            "Content-Disposition"
+        ] = f'attachment; filename="{settings.IMPORT_TEMPLATE_FILENAME}"'
+        utils.generate_import_template(response)
```
currycoder: I don't really follow how this works.  Doesn't `generate_import_template` expect a file path and not an HTTP response object?

app/enquiries/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    e.full_clean(["enquirer"])
+
+    # save and associate
+    enquirer.save()
+    e.enquirer = enquirer
+    e.save()
+    return e
+
+
+def generate_import_template(save_path):
```
currycoder: It would be helpful to have a docstring which details exactly what format of spreadsheet this will produce.

app/enquiries/tests/test_views.py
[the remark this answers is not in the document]
alixedi: I think what @mcbhenwood is saying here is that a test called `test_import_template` should test the import template e.g. whether the contents of the import template are valid and not just the `status_code`.

app/settings.py
```diff
@@ -133,3 +133,4 @@
 
 # App specific settings
 CHAR_FIELD_MAX_LENGTH = 255
+IMPORT_TEMPLATE_FILENAME = 'rtt_enquiries_import_template.xlsx'
```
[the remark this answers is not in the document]
alixedi: This does not seem consistent with the changelog: https://github.com/uktrade/enquiry-mgmt-tool/pull/18/files#diff-04c6e90faac2675aa89e2176d2eec7d8R66

Maybe I am missing something?

If the Excel spreadsheet is being generated dynamically, why do we have a management command to generate a `.xlsx` file?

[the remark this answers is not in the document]
alixedi: Answering my own Q, it seems the management command if for developers to generate the template locally.

I would suggest not putting this in production.

app/enquiries/tests/test_views.py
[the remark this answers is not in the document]
kiddhustle: Yes I haven't had a chance to look at this PR over the last few days but can look into add tests like this.

The SSO and filters PRs are a higher priority at the moment

app/enquiries/views.py
kiddhustle: the response has a file-like interface so we can write to it like we would a file

app/enquiries/utils.py
kiddhustle: Would you want this just for the enquiries sheet? The columns in the other sheets will be dynamically generated based on what is in the `ref_data` so maybe I could explain this in the docstring instead?

app/enquiries/tests/test_views.py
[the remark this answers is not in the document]
kiddhustle: @mcbhenwood @alixedi @currycoder

I have now added extra tests

app/enquiries/utils.py
currycoder: It sounds like both will be valuable :+1:

app/enquiries/views.py
currycoder: Ah I think it's the variable name `save_path` in `generate_import_template` which is tripping me up.  Could that possibly be renamed to something like `file_obj`?

kiddhustle: Yes sorry I think earlier it was a file path but since writing to the response the parameter has changed.

I will amend

app/enquiries/utils.py
kiddhustle: Added docs

app/enquiries/ref_data.py
```diff
@@ -411,3 +411,20 @@ class DatahubProjectStatus(models.TextChoices):
...
+    'enquirer_request_for_call',
+    'country',
+    'company_name',
+    'primary_sector',
+    'company_hq_address',
+    'website',
+    'investment_readiness',
+    'enquiry_text',
+    'notes'
+]
```
elcct: Would be good to fix newlines.

kiddhustle: amended

saruniitr:
Thanks for the reviews, received two approvals, merging it.

PR #22, opened 2020-03-12
Edit Enquiry

hnryjmes:
These are frontend and backend changes relating to editing and saving an enquiry.

Features:
- Edit button on Detail page
- Edit page with all inputs editable
- Data is pre-populated as expected
- Save, cancel buttons work as expected
- Full integration with Enquirer model

Screenshot:
screencapture-localhost-8000-enquiries-31-edit-2020-03-23-11_24_41


app/enquiries/migrations/0001_initial.py
```diff
@@ -1,10 +1,52 @@
...
+        {
+            "username": "ist-user-5",
+            "first_name": "IST",
+            "last_name": "User 5",
+            "email": "ist.user.5@example.com",
+        }
+    ]
+    for user in ist_users:
+        Owner.objects.create(user=User.objects.create(**user))
+
```
alixedi: Is this a fixture meant for testing?

alixedi:
Please clean up the commit history.

Each commit should be (1) atomic (2) contain a single change (3) with a commit message that summarises your intention.

This would make reviewing more pleasant and significantly quicker.

app/enquiries/migrations/0001_initial.py
hnryjmes: yes, but it disappears in future work - this dev work is around a month old, and we have moved on significantly as a project since then

hnryjmes: we can decide within the timescales of the project whether it is economical to refactor now, or to wait for it to disappear as soon as other PRs are merged

app/enquiries/migrations/0001_initial.py
```diff
@@ -169,7 +211,7 @@ class Migration(migrations.Migration):
                             ("LINKEDIN", "LinkedInLeadGen"),
                             ("CHOGM", "CHOGM"),
                             ("IIGB", "IiGB"),
-                            ("IIGB_LINKEDIN", "IiGB (LinkedIN)"),
+                            ("IIGB_LINKEDIN", "IiGB (LinkedIn)"),
```
currycoder: I've seen these changes across a few PRs - we need to be super careful that these are expunged from those PRs.  Migrations should not be changed once they are comitted.

app/enquiries/tests/test_views.py
```diff
@@ -215,6 +216,18 @@ def test_enquiry_successful_update(self):
...
+        enquirer = enquiry_obj.enquirer
+        data["enquirer"] = enquirer.id
+        data["first_name"] = enquirer.first_name
+        data["last_name"] = enquirer.last_name
+        data["job_title"] = enquirer.job_title
+        data["email"] = enquirer.email
+        data["phone"] = enquirer.phone
+        data["email_consent"] = enquirer.email_consent
+        data["phone_consent"] = enquirer.phone_consent
+        data["request_for_call"] = enquirer.request_for_call
```
currycoder: These have been set to make the test pass, but there's no additional assertions to ensure that enquirer details are updated.

app/enquiries/views.py
```diff
@@ -31,6 +32,14 @@ def get_paginated_response(self, data):
             }
         )
 
+    def post(self, request, format=None):
```
currycoder: I don't understand this.  Pagination objects don't have `post()` methods AFAIK.

app/enquiries/views.py
```diff
@@ -87,9 +96,15 @@ class EnquiryEditView(UpdateView):
     """
 
     model = models.Enquiry
-    fields = "__all__"
+    form_class = forms.EnquiryForm
     template_name = "enquiry_edit.html"
 
+    def get(self, request, *args, **kwargs):
```
currycoder: Is this overridden method necessary?  It seems to me like it's doing the same things that the original method would do.

app/enquiries/forms.py
```diff
@@ -0,0 +1,67 @@
...
+        self.fields['last_name'].initial = enquirer.last_name
+        self.fields['job_title'].initial = enquirer.job_title
+        self.fields['email'].initial = enquirer.email
+        self.fields['phone'].initial = enquirer.phone
+        self.fields['email_consent'].initial = enquirer.email_consent
+        self.fields['phone_consent'].initial = enquirer.phone_consent
+        self.fields['request_for_call'].initial = enquirer.request_for_call
+
+    def save(self, commit=True):
+        enquiry = super().save(commit=False)
```
currycoder: It would be safer to initiate a transaction than doing staged commits.  e.g.  we might end up with hanging enquirers if the eventual `enquiry.save()` fails for any reason.  Please see: https://github.com/uktrade/enquiry-mgmt-tool/blob/6469b3dca0cd99e6d364bc3663585aa550d0dcde/app/enquiries/serializers.py#L46

app/enquiries/migrations/0001_initial.py
saruniitr: As you mentioned earlier I have already fixed this. You are seeing these here because this was under review for a long time and `master` moved further.

app/enquiries/migrations/0001_initial.py
alixedi: Please make this a `pytest` fixture.

app/enquiries/tests/test_views.py
saruniitr: will look into and add any missing assertions

app/enquiries/views.py
saruniitr: no, looks it is because of auto merge conflict resolution, this should be in the UpdateView, will fix this

app/enquiries/views.py
saruniitr: Yes, it is necessary because we are creating a form instance and initializing `Enquirer` fields with initial value in `__init__` method. Original method was only initialising `enquirer.id` but we are using a single form to update all fields.

app/enquiries/migrations/0001_initial.py
saruniitr: I have already created a fixture. This branch is behind master as it was under review for a long time

app/enquiries/forms.py
alixedi: I presume this has been done because we will get the html representation of the form "for free" in the template. If this is correct, we are leaking presentation logic outside of templates. I would caution against that.

An alternative approach is to have 2 separate forms. One for the Enquirer and one for Enquiry. These could be fairly simple because `ModelForms`.

You can render a superset of the fields in whatever way you like in the template. Upon submission of the composite form, you can split the fields in the view function check `is_valid` for respective forms and if both are valid, save them in a `transaction.atomic()` as @currycoder has recommended ☝️

app/enquiries/tests/test_views.py
saruniitr: ok, there are no assertions for enquirer fields because they are not updated.
I have updated enquirer fields also and added assertions.

app/enquiries/views.py
saruniitr: removed

app/enquiries/forms.py
saruniitr: updated with atomic transaction.

hnryjmes:
I have now addressed the above comments:
- Rebased and squashed commit history following convention of one commit per working feature (with ticket number)
- Removed redundant Notes heading

frontend/sass/application.scss
```diff
@@ -111,5 +111,22 @@ a.entity-title-link {
 }
 
 .enquiry-details-row {
-    margin: govuk-spacing(8);
+    margin-bottom: govuk-spacing(8);
+}
+
+.govuk-fieldset__legend {
+    font-weight: bold;
+    margin-bottom: 5px;
```
ian-leggett: I would use `margin-bottom: govuk-spacing(1)`

hnryjmes:
thanks for the comment @ian-leggett, the change to the Notes heading has already been made per your previous comment 7 days ago - are you looking at the correct screenshot/most up to date version of the PR? :)

PR #23, opened 2020-03-12
Feature/RTT-62 sso

kiddhustle:
Adds Staff SSO integration

Details can be found at: 

https://uktrade.atlassian.net/browse/RTT-62


kiddhustle:
> I don't see tests that would show that this integration works.

@elcct when I spoke to Alessandro about how his team tests this integration he said that the tests of the library were sufficient and their app didn't test the integration itself. WOuld this be sufficient in this case?

app/settings.py
```diff
@@ -31,6 +32,9 @@
 
 ALLOWED_HOSTS = []
 
+FEATURE_FLAGS = {
+    'ENFORCE_STAFF_SSO_ON': env.bool('FEATURE_ENFORCE_STAFF_SSO_ENABLED', True),
```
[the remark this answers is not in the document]
kiddhustle: This is conditionally to aid with testing so that it can be disabled easily.

We set this env variable in `.circleci/config.yml`:

`FEATURE_ENFORCE_STAFF_SSO_ENABLED: 0`

app/settings.py
```diff
@@ -77,7 +81,30 @@
...
+
+# authbroker config
+if FEATURE_FLAGS['ENFORCE_STAFF_SSO_ON']:
+    INSTALLED_APPS.append('authbroker_client',)
+    AUTHBROKER_URL = env('AUTHBROKER_URL')
+    AUTHBROKER_CLIENT_ID = env('AUTHBROKER_CLIENT_ID')
+    AUTHBROKER_CLIENT_SECRET = env('AUTHBROKER_CLIENT_SECRET')
+
+    AUTHBROKER_TOKEN_SESSION_KEY = env('AUTHBROKER_TOKEN_SESSION_KEY')
+    # authentication
```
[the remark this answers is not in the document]
kiddhustle: formatting has been amended

kiddhustle:
> I don't see tests that would show that this integration works.

I have tried overriding the feature flag and whilst the setting is being overridden (when I print the setting value) it doesn't actually have an effect.

I have read online that if a module has beenloaded already (in this case the SSO Django client) the settings changes won't affect them.

So I am not sure how to go about testing at this point if I can't temporarily enable SSO for this test.

If I turn it on (via ENV variable/settings) then all other view tests will fail.

https://docs.djangoproject.com/en/3.0/topics/testing/tools/#overriding-settings

currycoder:
@kiddhustle To respond to your point about the difficulty of overriding these flags on the fly in tests, could it be better to flip things around a bit so that SSO auth is on by default?  I assume that this will be the case for nearly all environments (tests, staging, production) other than local development copies.

If that is true, then I'd be tempted to replace the feature flag setting implementation with a `DISABLE_SSO_AUTH` or similar flag which can be set for development environments.

kiddhustle:
> @kiddhustle To respond to your point about the difficulty of overriding these flags on the fly in tests, could it be better to flip things around a bit so that SSO auth is on by default? I assume that this will be the case for nearly all environments (tests, staging, production) other than local development copies.
> 
> If that is true, then I'd be tempted to replace the feature flag setting implementation with a `DISABLE_SSO_AUTH` or similar flag which can be set for development environments.

In this case would we have to implement mock SSO, if SSO was enabled by default?

@elcct did mention that he thought mock SSO would be overkill in our case. But if that's the only way I guess we will have to go with that

elcct:
@kiddhustle you could create a helper client for tests that will be making requests with already authenticated user, so that it will skip SSO authentication. You would also need then to add tests to at least check if user is not authenticated, that user is being redirected to whatever is set as `AUTHBROKER_URL` (https://github.com/uktrade/django-staff-sso-client/blob/master/authbroker_client/utils.py#L15).

You can see that SSO client middleware checks if user is authenticated here https://github.com/uktrade/django-staff-sso-client/blob/master/authbroker_client/middleware.py#L11 and if it is, then the request continues.

Given above, there is no need to use mock SSO

You can see an example how various clients that help running tests are constructed: https://github.com/uktrade/data-hub-api/blob/develop/datahub/core/test_utils.py

You could authenticate fake user for example here https://github.com/uktrade/enquiry-mgmt-tool/blob/master/app/enquiries/tests/test_views.py#L75

kiddhustle:
@elcct @currycoder @saruniitr I have done the following:

1. created a test customer
2. authenticated the user in the test client
3. logged the user out to test that they are redirect away from the list view

Let me know if this meets expectations

app/settings.py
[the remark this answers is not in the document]
kiddhustle: change to enable SSO by default

app/enquiries/tests/test_ping_views.py
```diff
@@ -9,6 +10,32 @@
 
 
 class ServiceHealthCheckTestCase(TestCase):
+    def setUp(self):
```
currycoder: Can you pop some newlines in between methods here and below?

app/enquiries/tests/test_views.py
```diff
@@ -260,3 +284,20 @@ def test_enquiry_detail_template_ref_data(self):
...
+        self.login()
+        self.logout()
+        self.assertEqual(self.logged_in, False)
+        # re-authenticate so subsequent tests can pass
+        self.login()
+    def test_login_protected(self):
+        """Test the view is protected by SSO"""
+        self.logout()
+        response = self.client.get(reverse("enquiry-list"))
+        self.assertNotEqual(response.status_code, status.HTTP_200_OK)
```
currycoder: I think it would be good to assert what this status code **is** and not that it "is not 200"

kiddhustle: have amended to check for a 302

app/enquiries/tests/test_ping_views.py
kiddhustle: reformatted

kiddhustle:
> I can't see a test for the bit that @elcct mentioned here:
> 
> > You would also need then to add tests to at least check if user is not authenticated, that user is being redirected to whatever is set as AUTHBROKER_URL (https://github.com/uktrade/django-staff-sso-client/blob/master/authbroker_client/utils.py#L15).

now checks the URL

app/enquiries/tests/test_views.py
```diff
@@ -68,12 +69,35 @@ def canned_enquiry():
...
 class EnquiryViewTestCase(TestCase):
     def setUp(self):
+        """
+        Sets up test user to authenticate requests.
+        If you need to log out for a test remember to log bck in afterwards
+        """
         self.faker = Faker()
         self.client = Client()
 
+        self.CREDENTIALS = {
```
elcct: There is some repetition there. It would be good to create a helper class so that it can be reused instead of setting up the test user separately in each one.

app/enquiries/tests/test_views.py
```diff
@@ -260,3 +284,23 @@ def test_enquiry_detail_template_ref_data(self):
         country_display_name = get_display_name(ref_data.Country, enquiry.country)
         self.assertContains(response, enquiry_stage_display_name)
         self.assertContains(response, country_display_name)
+    def test_helper_login(self):
+        result = self.login()
+        self.assertEqual(result, True)
```
elcct: I am not sure if this test is useful - it doesn't actually check that the client has logged in user.

app/enquiries/tests/test_ping_views.py
```diff
@@ -9,6 +10,32 @@
...
+        UM = get_user_model()
+        # create test user
+        u = UM(**self.CREDENTIALS)
+        u.set_password(self.CREDENTIALS.get("password"))
+        u.save()
+        # login
+        self.login()
+
+    def login(self):
+        self.logged_in = self.client.login(**self.CREDENTIALS)
```
elcct: the `self.logged_in` doesn't seem to be used anywhere.

app/enquiries/tests/test_views.py
kiddhustle: @elcct Yes I was actually thinking the same thing but I didn't actually have to do it.

app/enquiries/tests/test_ping_views.py
reupen: It looks like this is the health check endpoint, in which case requiring user authentication is probably going to be problematic. Is there a reason for it?

app/enquiries/tests/test_views.py
```diff
@@ -68,12 +69,35 @@ def canned_enquiry():
...
+        """
         self.faker = Faker()
         self.client = Client()
 
+        self.CREDENTIALS = {
+            "username": "test",
+            "password": "user"
+        }
+        self.logged_in = False
+        UM = get_user_model()
```
reupen: Would be good to use a more descriptive name than `UM` e.g. `user_model`:
```suggestion
        user_model = get_user_model()
```

app/enquiries/tests/test_views.py
```diff
@@ -68,12 +69,35 @@ def canned_enquiry():
...
+        # create test user
+        u = UM(**self.CREDENTIALS)
```
reupen: This is passing `password` (a hash) to the user object and then overwriting it below. So would suggest:
```suggestion
        user = UM(usename=self.CREDENTIALS['username'])
```

app/enquiries/tests/test_views.py
```diff
@@ -68,12 +69,35 @@ def canned_enquiry():
...
+        u.set_password(self.CREDENTIALS.get('password'))
```
reupen: `dict.get()` is for optional keys – so would suggest using `[` and `]`:
```suggestion
        user.set_password(self.CREDENTIALS['password''])
```

app/enquiries/tests/test_views.py
reupen: I agree that it's not clear what is being tested. Perhaps you could clarify?

app/enquiries/tests/test_views.py
```diff
@@ -260,3 +284,23 @@ def test_enquiry_detail_template_ref_data(self):
...
         self.assertContains(response, country_display_name)
+    def test_helper_login(self):
+        result = self.login()
+        self.assertEqual(result, True)
+        self.assertEqual(self.logged_in, True)
+        self.assertEqual(result, self.logged_in)
+    def test_helper_logout(self):
+        self.login()
+        self.logout()
+        self.assertEqual(self.logged_in, False)
```
reupen: I'm not sure what this is testing: `self.logout()` unconditionally sets `self.logged_in` to False.

app/enquiries/tests/test_views.py
kiddhustle: but by returning true does that show that login was successfully

app/enquiries/tests/test_ping_views.py
kiddhustle: not in this test but amended with latest commit

app/enquiries/utils.py
```diff
@@ -0,0 +1,4 @@
+from django.conf import settings
+
+def get_oauth_payload(request):
+    return request.session.get(settings.AUTHBROKER_TOKEN_SESSION_KEY)
```
reupen: Could you add a docstring to explain what does does? And should it be using `[` and `]` rather than `.get()`?

app/enquiries/tests/test_views.py
kiddhustle: Now refactored

app/enquiries/tests/test_views.py
```diff
@@ -4,8 +4,9 @@
 
 from datetime import date
 from django.conf import settings
+from django.contrib.auth import get_user_model
 from django.forms.models import model_to_dict
-from django.test import Client, TestCase
+from django.test import Client, TestCase, override_settings
```
reupen: `override_settings` doesn't seem to be used. It looks like at least one of the tests depends on SSO being turned on, so it should probably be used for that.

app/enquiries/tests/test_ping_views.py
kiddhustle: This was recently merged in from `master` and I added it for consistency.

So this should be a publicly available endpoint?

app/enquiries/tests/test_views.py
kiddhustle: This is from previous attempts to test SSO.

 override_settings doesn't work as expected in our case so know we have SSO enabled by default for tests. hence why we have the login/logout methods on the base test case class (in latest commit)

app/enquiries/tests/test_ping_views.py
reupen: The health check endpoint will need to be accessible from at least Pingdom – most of the health check endpoints are public I believe, yes.

app/enquiries/utils.py
kiddhustle: I think this one should be a `.get` I have added a fallback `None`

app/enquiries/tests/test_views.py
kiddhustle: now removed

app/enquiries/tests/test_views.py
kiddhustle: added extra test to check that login sets to `True`, the logout sets to `False`.

Does this now make sense?

app/enquiries/tests/test_views.py
reupen: I can see `FEATURE_ENFORCE_STAFF_SSO_ENABLED` is set in the CircleCI config but is it being set anywhere else? The tests shouldn't really depend on environment variables in that manner (they shouldn't fail if you have that turned off locally, for example).

app/enquiries/models.py
```diff
@@ -30,15 +30,13 @@ class Enquirer(models.Model):
     )
 
 
-class Owner(models.Model):
+class Owner(AbstractUser):
     """
     Model for the user assigned to an Enquiry
```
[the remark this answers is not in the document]
kiddhustle: docs added

app/enquiries/tests/test_ping_views.py
kiddhustle: ok I will make these public

app/enquiries/tests/test_views.py
kiddhustle: added

app/urls.py
```diff
@@ -13,20 +13,26 @@
...
     path('enquiries/', views.EnquiryListView.as_view(), name="enquiry-list"),
     path('enquiries/<int:pk>/', views.EnquiryDetailView.as_view(), name="enquiry-detail"),
     path('enquiries/<int:pk>/edit', views.EnquiryEditView.as_view(), name="enquiry-edit"),
     path('healthcheck/ping', ping.ping, name='ping'),
 ]
+
+if settings.FEATURE_FLAGS["ENFORCE_STAFF_SSO_ON"]:
+    urlpatterns.append(
+        path('auth/', include('authbroker_client.urls')),
+    )
```
elcct: It would be good to add newline.

kiddhustle: formatting amended

app/settings/common.py
```diff
@@ -138,26 +165,20 @@
...
+    {"NAME": "django.contrib.auth.password_validation.CommonPasswordValidator",},
+    {"NAME": "django.contrib.auth.password_validation.NumericPasswordValidator",},
 ]
 
 
 # Internationalization
 # https://docs.djangoproject.com/en/3.0/topics/i18n/
 
-LANGUAGE_CODE = 'en-us'
+LANGUAGE_CODE = "en-us"
```
elcct: That should be `en-gb`, no?

PR #26, opened 2020-03-16
Pagination

hnryjmes:
Frontend changes to add pagination, including:
- basic logic to hide/show buttons in the template
- CSS styling to match Data Hub pagination

Screenshot:

Screenshot 2020-03-16 at 15 29 11

### DIFF
--- a/app/enquiries/templates/enquiry_list.html
+++ b/app/enquiries/templates/enquiry_list.html
@@ -73,7 +73,39 @@
                 </li>
                 {% endfor %}
             </ol>
-            <nav class="pagination"></nav>
+
+            <hr class="govuk-section-break govuk-section-break--m govuk-section-break">
+
+            <nav class="pagination">
+
+                <ol class="pagination-list">
+                    {%if previous is not None %}
+                    <li class="pagination-list-item">
+                        <a class="pagination-link" href="{{ previous }}">Previous</a>
+                    </li>
+                    {%endif%}
+
+                    {% for page in page_range %}
+
+                    {% if page|stringformat:"s" == current_page %}
+
+                    <a class="pagination-link" href="{% url 'enquiry-list' %}?page={{ page }}"><li class="pagination-list-item pagination-list-item-current">{{ page }}</li></a>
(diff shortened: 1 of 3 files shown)


app/enquiries/templates/enquiry_list.html
```diff
@@ -73,7 +73,39 @@ <h3>
...
+                <ol class="pagination-list">
+                    {%if previous is not None %}
+                    <li class="pagination-list-item">
+                        <a class="pagination-link" href="{{ previous }}">Previous</a>
+                    </li>
+                    {%endif%}
+
+                    {% for page in page_range %}
+
+                    {% if page|stringformat:"s" == current_page %}
```
currycoder: I think it would be nice if the template code was indented in for/if blocks.

app/enquiries/tests/test_views.py
```diff
@@ -70,7 +70,7 @@ def canned_enquiry():
 
 REST_FRAMEWORK_TEST = {
     "DEFAULT_PAGINATION_CLASS": "rest_framework.pagination.PageNumberPagination",
-    "PAGE_SIZE": 2,
+    "PAGE_SIZE": 10,
```
currycoder: Why have these values been adjusted?  I believe that the original idea behind keeping these low was to easily test the edge cases around pagination without having to create a lot of data per test.  e.g. create enough entities for a first full page of items and a second page with just one record on it.

hnryjmes: @saruniitr see above, the reason for this was a bug we saw in development, can you give more information?

saruniitr: Yes this was a temporary change to unblock frontend. The fixes are done in backend and they were under review.
Now they are merged and this doesn't exist in `master`. Updating with `master` will fix this.

saruniitr: Reverted temporary fix and updated with `master`.

saruniitr:
Thanks for reviews, two approvals received, merging it.

PR #27, opened 2020-03-20
Changes for deploying the Application

saruniitr:
## Change description

This PR adds changes to deploy the application to Paas Environment. In this case initially dev environment is chosen.

The app is currently deployed to dev environment using this changes and is available at,
https://enquiry-mgmt-dev.london.cloudapps.digital/


app/settings/local.py
```diff
@@ -0,0 +1,15 @@
...
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.postgresql',
+        'NAME': env('POSTGRES_DB_NAME', default='postgres'),
+        'USER': env('POSTGRES_DB_USER', default='postgres'),
+        'PASSWORD': env('POSTGRES_DB_PASSWORD'),
+        'HOST': env('POSTGRES_DB_HOST'),
+        'PORT': env.int('POSTGRES_DB_PORT', default=5432),
+    }
+}
```
reupen: Could local use `DATABASE_URL` as well?

manifest.yml
```diff
@@ -0,0 +1,6 @@
+---
+applications:
+-
+  buildpack:
```
reupen: Is the newline after `-` needed?
```suggestion
- buildpack:
```

web.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# Start-up script for the web process, primarily for GOV.UK PaaS (see the Procfile)
+
+set  -xe
+
+# Compile css
+npm rebuild node-sass && npm install --only=dev && npm run sass
+
+./manage.py migrate --noinput
+./manage.py collectstatic --noinput
```
reupen: Normally the buildpack runs `collectstatic` (unless `DISABLE_COLLECTSTATIC` is set) – was that not working?

app/settings/local.py
saruniitr: yes but I still need to provide host, user etc so thought its easier to keep it like this.

manifest.yml
saruniitr: not really, initially some syntax errors were causing weird issues so just followed as in example. Will remove it if not required.

web.sh
saruniitr: It is really not necessary as the `gulp` script is placing the files in correct directory, just added this for completeness as it shouldn't have any effect, will remove it.

manifest.yml
saruniitr: fixed

app/settings/dev.py
```diff
@@ -0,0 +1 @@
+from app.settings.common import *
```
reupen: Is this settings module really needed?

app/settings/local.py
```diff
@@ -0,0 +1 @@
+from app.settings.common import *
```
reupen: Is this settings module really needed?

app/settings/dev.py
saruniitr: well it is not required at the moment but left it for future requirement.
I guess I can add it back when required, will remove it now.

app/settings/local.py
saruniitr: same as above

reupen:
> Could you clarify what you mean tidy up commits?

Usually we squash unnecessary commits before merging. In this case, there are 31 commits, but it could probably be somewhere between 1 and 3 commits. So, I mean, once you have a second approval, it would be good to rebase and squash some of the commits.

saruniitr:
IMHO I think it is better to leave as is, if we squash we lose history. The large number of commits in branch is main because we can only test the changes after deployment. If squashing is preferred I can merge using "Squash and merge" option.

saruniitr:
@alixedi I have addressed all comments, waiting for the second approval to merge.

saruniitr:
Thanks for your reviews, received two approvals, merging it.

PR #28, opened 2020-03-25
Submit Investment data to Data Hub

saruniitr:
## Change description

This PR adds changes to submit investment enquiry data to Data Hub.

During this process it first checks whether a company exists in Data Hub otherwise it returns an error (see below). Next we check for any existing contacts and add the contact as secondary/primary. It also searches for the assigned adviser and finally submits the data.

At any point if there is an error, the list of errors till that point are presented to the user (as per example screenshot below).

The relevant metadata is fetched and mapped to the unique ids as required by investment creation API. There is a mismatch of reference data coming from the website as compared to what is available in Data Hub, in this case we won't get a uuid and treated as error.

SSO is now integrated so there are few changes to remove previous TODOs.

Company not available in Data Hub
company-not-in-dh


app/enquiries/common/datahub_utils.py
```diff
@@ -213,3 +222,205 @@ def dh_contact_search(contact_name):
...
+    """
+    url = settings.DATA_HUB_CONTACT_CREATE_URL
+    enquirer = enquirer.enquirer
+    payload = {
+        "first_name": enquirer.first_name,
+        "last_name": enquirer.last_name,
+        "job_title": enquirer.job_title,
+        "company": company_id,
+        "primary": primary,
+        "telephone_countrycode": "NOT SET",
```
reupen: This should be a real value rather than NOT SET. So either the phone number needs to be captured from the user as two fields, or it needs splitting afterwards.

app/enquiries/common/datahub_utils.py
```diff
@@ -213,3 +222,205 @@ def dh_contact_search(contact_name):
...
+def dh_adviser_search(request, access_token, adviser_name):
+    """
+    Peforms an Adviser search using Data hub API.
+
+    Returns list of subset of fields for each Adviser found
+    """
+    advisers = []
+    url = f"{settings.DATA_HUB_ADVISER_SEARCH_URL}/?autocomplete={adviser_name}"
+
+    response = dh_request(request, access_token, "GET", url, {})
```
reupen: It would be good to pass query params to requests using the `params` keyword argument so that they're escaped correctly.

app/enquiries/common/datahub_utils.py
```diff
@@ -213,3 +222,205 @@ def dh_contact_search(contact_name):
...
+        return advisers, response.json()
+
+    for adviser in response.json()["results"]:
+        advisers.append(
+            {
+                "datahub_id": adviser["id"],
+                "name": adviser["first_name"],
+                "is_active": adviser["is_active"],
+            }
+        )
```
reupen: It would be good to use a list comprehension here e.g.:

```python
advisers = [
    {
        "datahub_id": adviser["id"],
        "name": adviser["first_name"],
        "is_active": adviser["is_active"],
    }
    for adviser in response.json()["results"]
]
```

It's both idiomatic and more efficient.

app/enquiries/common/datahub_utils.py
```diff
@@ -213,3 +222,205 @@ def dh_contact_search(contact_name):
...
+        response["errors"].append(
+            {
+                "enquiry": f"Enquiry can only be submitted once,"
+                f" previously submitted on {prev_submission_date}, stage {stage}"
+            }
+        )
+        return response
+
+    # check if the user is available in Data Hub
+    advisers, error = dh_adviser_search(request, access_token, request.user)
```
reupen: Is this trying to get the adviser associated with the access token? The usual way to do this is to make a request to `/whoami/`.

app/enquiries/common/datahub_utils.py
```diff
@@ -213,3 +222,205 @@ def dh_contact_search(contact_name):
...
+        return response
+
+    payload["client_relationship_manager"] = advisers[0]["datahub_id"]
+
+    payload["sector"] = map_to_datahub_id(
+        enquiry.get_primary_sector_display(), dh_metadata, "sector"
+    )
+
+    # It is always default value - corresponds to Services
+    payload["business_activities"] = ["2f51ea6a-ca2f-466a-87fd-5f79ebfec125"]
```
reupen: It would be good to make this UUID a named constant.

app/enquiries/common/datahub_utils.py
```diff
@@ -213,3 +222,205 @@ def dh_contact_search(contact_name):
...
+    payload["referral_source_adviser"] = referral_adviser
+    payload["referral_source_activity"] = get_dh_id(
+        dh_metadata["referral-source-activity"], "Website"
+    )
+    payload["referral_source_activity_website"] = get_dh_id(
+        dh_metadata["referral-source-website"], "Invest in GREAT Britain"
+    )
```
reupen: It would be good to be consistent about whether these are referenced by name or by ID. I would suggest ID (possibly even overridable in setting) as they can be freely renamed from the admin site while the UUIDs are fairly persistent.

app/enquiries/common/datahub_utils.py
```diff
@@ -213,3 +222,205 @@ def dh_contact_search(contact_name):
...
+    return advisers, None
+
+
+def get_dh_id(metadata_items, name):
+    item = list(filter(lambda x: x["name"] == name, metadata_items))
+    assert len(item) == 1
+    return item[0]["id"]
+
+
+def dh_investment_create(request, enquiry, metadata=None):
```
reupen: This function is quite large – it would be good to split it up.

app/settings/common.py
```diff
@@ -49,7 +49,7 @@
     },
     'loggers': {
         '': {
-            'level': 'DEBUG' if env.bool('DEBUG') else 'INFO',
+            'level': 'INFO' if env.bool('DEBUG') else 'INFO',
```
reupen: Can this be just 'INFO' now?
```suggestion
            'level': 'INFO',
```

app/enquiries/tests/test_dh_integration.py
```diff
@@ -161,6 +174,41 @@ def test_contact_search_error(self):
...
             expected = contact_search_response()["error"]
 
-            response, error = dh_contact_search("")
+            response, error = dh_contact_search(
+                "mock_request", "access_token", "", "company_id"
+            )
             self.assertIsNotNone(error)
             self.assertEqual(error["name"], expected["name"])
+
+    def test_investment_creation_fails_company_not_in_dh(self):
```
reupen: It there a test for when `dh_investment_create()` succeeds? It would be good to check that the expected data is being sent to Data Hub.

app/enquiries/common/datahub_utils.py
```diff
@@ -213,3 +222,205 @@ def dh_contact_search(contact_name):
...
+    If this is the only contact then it will be made primary.
+    """
+
+    # Return a list of errors to be displayed in UI
+    response = {
+        "errors": []
+    }
+
+    # Extract access token
+    session = get_oauth_payload(request)
```
elcct: I think there is a duplication of this logic. Why the token needs to be passed explicitly if it is being extracted in `dh_request` if not provided? It would be good to clean this up.

app/enquiries/common/datahub_utils.py
elcct: If you are trying to find user by name, be aware that there are cases where advisers share the same name. I think what @reupen  suggests is the way to go.

app/enquiries/models.py
```diff
@@ -216,6 +214,40 @@ class Enquiry(TimeStampedModel):
...
+        null=True,
+        verbose_name="Company number in Data Hub",
+    )
+    dh_duns_number = models.CharField(
+        max_length=MAX_LENGTH,
+        blank=True,
+        null=True,
+        verbose_name="Duns number",
+    )
+    dh_assigned_company = models.CharField(
```
elcct: It would be good to rename it to `dh_assigned_company_name`, as otherwise it suggests the column may store the company object, not just the name.

app/enquiries/common/datahub_utils.py
saruniitr: Unfortunately the form where user enters this data has one column, no field to capture the code so we cannot rely on the number to split it up later. We have a backlog item to add a field in the tool then service team can enter the code based on the country. I was planning to do a look up but I think I make the form changes in our tool itself to enter the country code.

app/enquiries/common/datahub_utils.py
saruniitr: ok

app/enquiries/common/datahub_utils.py
saruniitr: I was favouring readability but I can change it to use list comprehension.

app/enquiries/common/datahub_utils.py
saruniitr: Yes this is find the user on Data Hub as they will be recorded as referral source adviser. I will update this to make use of `/whoami/`

app/enquiries/common/datahub_utils.py
saruniitr: agree. Initially I was thinking of importing data hub fixtures and load it up from there but it can quickly go out of sync and adds maintenance. Since this is a fixed field I have used the id directly but will make it named for readability.

app/enquiries/common/datahub_utils.py
saruniitr: ok, there are only few fields that require this. Plan was to refine this in later releases.

app/enquiries/common/datahub_utils.py
saruniitr: I mainly included in a single function to handle errors easily and to collect error descriptions as we need to display them in the UI. I will see if it can be refactored.

app/settings/common.py
saruniitr: yes, finding DEBUG too verbose, will make it INFO.

app/enquiries/tests/test_dh_integration.py
saruniitr: Not at the moment, adding it.

app/enquiries/common/datahub_utils.py
saruniitr: I am passing both `request` and `access_token` to `dh_request` to have some flexibility as it is used by other functions (company/contact/adviser search, contract create and investment create).
Only investment create receives `request` object which calls other functions but there is a case where company search is called separately. To cover these case I am sending token separately. I agree it causes some duplication but favoured flexibility to support all cases.

app/enquiries/models.py
saruniitr: hmm, will change it to make it clear.

app/enquiries/common/datahub_utils.py
saruniitr: Also in future these may come from the enquiry model itself, for now I have changed them as defines.

saruniitr:
@reupen @elcct @mcbhenwood  With reference to tests, my limitation is that I can only mock requests to datahub as the DH API is not running locally. CI cannot really access data hub so no option but to mock responses. This means I need to setup lot of mock responses, data hub metadata in this project and needs to be updated as and when API changes. Do you have any recommendation on how to resolve this?

reupen:
I think that is right for unit-ish tests. We don't have any kind of mock client available in Python or any other language, so it would be requests-mock I think.

For end-to-end tests, you could use the data-hub-api Docker image with its fixtures. But I don't believe there are any end-to-end tests in this repo at the moment.

(There is also something called that sandbox in the Data Hub FE, but that's a bunch of fixtures and a fake API server.)

In relation to metadata lists, there are two types – ones that are managed in the Data Hub code base and ones that are editable from the admin site. For the second group, fewer assumptions can be made.

reupen:
@saruniitr Let me know when this is ready for re-review.

saruniitr:
> @saruniitr Let me know when this is ready for re-review.

Addressed previous comments and refactored it. Once the tests are complete will let you know.

saruniitr:
Have address previous review comments.
Also integrated cypress test framework and started adding a test for investment creation using sandbox api but the `POST` request on `/v3/investment` giving `route invalid` error. Hence the changes are not committed to this PR. Checking with FE team about that error, if that is resolved then I will push those changes.

.circleci/config.yml
```diff
@@ -18,6 +18,10 @@ jobs:
       DATA_HUB_METADATA_URL: 'https://datahub/metadata'
       DATA_HUB_COMPANY_SEARCH_URL: 'https://datahub/company/search'
       DATA_HUB_CONTACT_SEARCH_URL: 'https://datahub/contact/search'
+      DATA_HUB_CONTACT_CREATE_URL: 'https://datahub/contact/create'
+      DATA_HUB_INVESTMENT_CREATE_URL: 'https://datahub/investment/create'
+      DATA_HUB_ADVISER_SEARCH_URL: 'https://adviser/search'
+      DATA_HUB_WHOAMI: 'https://datahub/whoami/'
```
elcct, line 24: It is missing `_URL`.

app/enquiries/common/datahub_utils.py
```diff
@@ -24,7 +28,17 @@
...
-def dh_request(method, url, payload, request_headers=None, timeout=15):
+
+def dh_request(
+    request,
+    access_token,
+    method,
+    url,
+    payload,
+    request_headers=None,
+    params={},
```
elcct, line 39: It's probably better to have `None` as default for `params` and then assign empty dictionary inside the function.

.circleci/config.yml
saruniitr: fixed

app/enquiries/common/datahub_utils.py
saruniitr: yes agree, as it is mutable it will cause issues with default arguments, corrected it.

app/enquiries/tests/test_dh_integration.py
elcct: Has this test been added?

app/enquiries/templates/enquiry_detail.html
```diff
@@ -18,6 +18,27 @@
...
+            </h2>
+            {% for error in errors %}
+                <div class="govuk-error-summary__body">
+                <ul class="govuk-list govuk-error-summary__list">
+                    {% for key,value in error.items %}
+                    <li> {{ key.capitalize }}: {{ value}} </li>
+                    {% endfor %}
+                </ul>
+                </div>
+            {% endfor %}
```
peterhudec, line 34: The outer loop looks weird, there surely should only be a single `.govuk-error-summary__body`. I would expect that the for loop in the `<ul>` would iterate over the errors. What are `error.items`?

app/enquiries/templates/enquiry_detail.html
```diff
@@ -105,31 +127,38 @@ <h3 class="govuk-heading-m">HPO details</h3>
...
+            <h3 class="govuk-heading-m">Notes</h3>
+            <dl class="govuk-summary-list">
+                {% include 'snippets/summary_list_key_value.html' with instance=enquiry field="notes" %}
+            </dl>
+        </div>
+    </main>
+    <form class="form" action="{% url 'enquiry-detail' pk=enquiry.id %}" method="post">
+        {% csrf_token %}
+
+        {% if not enquiry.date_added_to_datahub %}
```
peterhudec, line 159: Shouldn't the whole form be inside of this condition?

app/enquiries/common/datahub_utils.py
```diff
@@ -187,31 +221,310 @@ def dh_company_search(company_name):
...
+
+
+def dh_adviser_search(request, access_token, adviser_name):
+    """
+    Peforms an Adviser search using Data hub API.
+
+    Returns list of subset of fields for each Adviser found
+    """
+    advisers = []
+    url = f"{settings.DATA_HUB_ADVISER_SEARCH_URL}"
```
reupen, line 288: Looks like this f-string could be removed:

```suggestion
    url = settings.DATA_HUB_ADVISER_SEARCH_URL
```

app/enquiries/tests/test_dh_integration.py
saruniitr: No, sandbox api doesn't support this yet so could not add it.

app/enquiries/templates/enquiry_detail.html
saruniitr: It can be but this is a static page and there are no input fields, just a button at the bottom of a lengthy page to submit, for this reason only enclosed the button in the form tag.

app/enquiries/templates/enquiry_detail.html
saruniitr: Here errors is a dict object, these are not form errors. In the backend a dictionary is constructed with keys as error field and value as description of the error. Please see dh_enquiry_readiness() in datahub_utils.py. An example is shown below,
```
response = {"errors": []}
# Allow creating of investments only if Company exists on DH
if not enquiry.dh_company_id:
    response["errors"].append(
        {"company": f"{enquiry.company_name} doesn't exist in Data Hub"}
...
....
```

web-bert: I agree with @peterhudec also the `<li>` should have a link in it to the element in the error state. If the errors are not in a format that works for the template then they should be normalised to work for the template. I would also argue that the error summary should be a reusable snippet that takes a list of errors with a message and an id to link to the element in the error state

app/enquiries/templates/enquiry_detail.html
```diff
@@ -105,31 +127,38 @@ <h3 class="govuk-heading-m">HPO details</h3>
...
+                {% include 'snippets/summary_list_key_value.html' with instance=enquiry field="estimated_land_date" %}
+                {% include 'snippets/summary_list_key_value.html' with instance=enquiry field="new_existing_investor" %}
+                {% include 'snippets/summary_list_key_value.html' with instance=enquiry field="investor_involvement_level" %}
+                {% include 'snippets/summary_list_key_value.html' with instance=enquiry field="specific_investment_programme" %}
+                {% include 'snippets/summary_list_key_value.html' with instance=enquiry field="crm" %}
+                {% include 'snippets/summary_list_key_value.html' with instance=enquiry field="date_added_to_datahub" %}
+                {% include 'snippets/summary_list_key_value.html' with instance=enquiry field="project_code" %}
+                {% include 'snippets/summary_list_key_value.html' with instance=enquiry field="datahub_project_status" %}
+                {% include 'snippets/summary_list_key_value.html' with instance=enquiry field="project_success_date" %}
+            </dl>
```
web-bert, line 147: This could be more DRY by creating a reusable `summary-list` snippet?

saruniitr:
Thanks for your reviews, received three approvals, merging it.

PR #29, opened 2020-03-26
Fetch enquiries from AS

saruniitr:
## Change description

This PR adds changes to periodically fetch new investment enquiries data from Activity stream (AS) by means of a celery task. The default interval of polling is set to 30 min.
The enquiry details are present in the html/text body of the data so for each item we parse the body and extract relevant data.
Certain reference data as required by the form in the website is different from the dit-sectors reference data (which is the reference for this project also). Because of this some of the fields need to be mapped to the project.
I am currently adding tests.


app/settings/common.py
```diff
@@ -252,3 +252,16 @@
...
+ACTIVITY_STREAM_KEY_ID = env('ACTIVITY_STREAM_KEY_ID')
+ACTIVITY_STREAM_KEY = env('ACTIVITY_STREAM_KEY')
+ACTIVITY_STREAM_INITIAL_LOAD_DATE = env('ACTIVITY_STREAM_INITIAL_LOAD_DATE')
+ACTIVITY_STREAM_SEARCH_URL = env('ACTIVITY_STREAM_SEARCH_URL')
+ACTIVITY_STREAM_SEARCH_TARGET_URL = env('ACTIVITY_STREAM_SEARCH_TARGET_URL')
+ACTIVITY_STREAM_ENQUIRY_SEARCH_KEY1 = env('ACTIVITY_STREAM_ENQUIRY_SEARCH_KEY1')
+ACTIVITY_STREAM_ENQUIRY_SEARCH_VALUE1 = env('ACTIVITY_STREAM_ENQUIRY_SEARCH_VALUE1')
+ACTIVITY_STREAM_ENQUIRY_SEARCH_KEY2 = env('ACTIVITY_STREAM_ENQUIRY_SEARCH_KEY2')
+ACTIVITY_STREAM_ENQUIRY_SEARCH_VALUE2 = env('ACTIVITY_STREAM_ENQUIRY_SEARCH_VALUE2')
+ACTIVITY_STREAM_ENQUIRY_DATA_OBJ = env('ACTIVITY_STREAM_ENQUIRY_DATA_OBJ')
```
reupen: It would be good to document what these settings are for.

requirements.txt
```diff
@@ -1,5 +1,6 @@
 attrs==19.3.0
 backcall==0.1.0
+bs4==0.0.1
```
reupen: This is some kind of stub package – the proper one is `beautifulsoup4`.

app/enquiries/common/as_utils.py
```diff
@@ -0,0 +1,277 @@
...
+    Parses email body in the submission and returns Enquiry related fields.
+    This can be used to create an Enquiry.
+    """
+    enquiry_data = {}
+
+    body = submission["_source"]["object"][settings.ACTIVITY_STREAM_ENQUIRY_DATA_OBJ][
+        "html_body"
+    ]
+    if not body:
+        return enquiry_data
```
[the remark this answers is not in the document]
saruniitr: unlikely, additional guarding just in case the data key is missing or different. May adding an error log is better if this is the case or remove it completely?

app/enquiries/common/as_utils.py
```diff
@@ -0,0 +1,277 @@
...
+    enquiry["enquirer"]["first_name"] = data["Given name"]
+    enquiry["enquirer"]["last_name"] = data["Family name"]
+    enquiry["enquirer"]["job_title"] = data["Job title"]
+    if data.get("Email address"):
+        email = data["Email address"]
+    elif data.get("Work email address"):
+        email = data["Work email address"]
+    enquiry["enquirer"]["email"] = email
+    enquiry["enquirer"]["phone"] = data["Phone number"]
+    true_or_false = lambda value: True if value == "True" else False
```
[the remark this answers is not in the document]
saruniitr: ah of course, `value == "True"` already results in bool, will correct this

app/enquiries/common/as_utils.py
```diff
@@ -0,0 +1,277 @@
...
+
+    """
+    The query parameters are not sufficient to filter investment enquiries
+    hence using url as additional filter but some of the other enquiries share
+    the same url hence additionally look for specified keys to filter data
+    """
+    if not (
+        set(["Given name", "Job title", "Company HQ address"])
+        <= set(enquiry_data.keys())
+    ):
```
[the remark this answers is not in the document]
saruniitr: just reduces the overhead of creating a set, will correct this.

app/settings/common.py
saruniitr: ok

requirements.txt
saruniitr: ok

saruniitr:
Thanks for your reviews, received two approvals, merging it

PR #30, opened 2020-03-27
Delete Enquiry

hnryjmes:
This sets up a view to handle deleting an enquiry plus template and style changes.

The user has to click twice to delete an enquiry, once to initiate, once to confirm.

Screenshots:
Screenshot 2020-03-30 at 09 10 14
Screenshot 2020-03-30 at 09 10 23


frontend/sass/application.scss
```diff
@@ -165,4 +165,16 @@ li.pagination-list-item-current, li:hover.pagination-list-item-current {
...
+    float: right;
+}
+
+form.delete-form {
+    display: inline;
+}
+
+.delete-warning {
+    display: inline-block;
 }
```
ian-leggett: As I mentioned before this is file is just growing with CSS tacked on at the end, also we should follow BEM style and not hook classes to HTML elements like `form.delete-form`. Can this CSS be split up and organised into separate files?

app/enquiries/views.py
```diff
@@ -110,3 +110,17 @@ def form_invalid(self, form):
...
+    """
+
+    model = models.Enquiry
+    template_name = "enquiry_delete.html"
+
+    def post(self, request, **kwargs):
+        pk = kwargs["pk"]
+        enquiry = get_object_or_404(models.Enquiry, pk=kwargs["pk"])
+        enquiry.delete()
+        return redirect("enquiry-list")
```
saruniitr: please newline endings for this file and the css file below

PR #31, opened 2020-03-27
Feature/rtt 16 export enquiries backend

kiddhustle:
add the CSV export view.

The user requests this endpoint to export all enquiries in CSV format

### DIFF
--- a/app/enquiries/tests/test_utils.py
+++ b/app/enquiries/tests/test_utils.py
@@ -1,3 +1,7 @@
+import codecs
+import csv
+import io
+
 
 from django.test import Client, TestCase
 from faker import Faker
@@ -6,6 +10,7 @@ import app.enquiries.tests.utils as test_utils
 from app.enquiries import models, ref_data, utils
 
 from app.enquiries.tests.factories import (
+    EnquiryFactory,
     create_fake_enquiry_csv_row,
     get_random_item,
     get_display_name,
@@ -50,7 +55,6 @@ class EnquiryViewTestCase(TestCase):
         ).exists()
         self.assertTrue(exists)
 
-
 class UtilsTestCase(test_utils.BaseEnquiryTestCase):
     def test_helper_logout(self):
         """
@@ -62,3 +66,38 @@ class UtilsTestCase(test_utils.BaseEnquiryTestCase):
         self.assertEqual(self.logged_in, False)
         # re-authenticate so subsequent tests can pass
         self.login()
+
+    def test_util_export_to_csv(self):
+        """
+        Tests that the export to csv util writes the expected data to file
+        """
+        enquiries = [EnquiryFactory() for i in range(5)]
+        qs = models.Enquiry.objects.all()
+        fp = io.StringIO()
+
+        utils.export_to_csv(qs, fp)
+        reader = csv.DictReader(fp)
+        row_count = 0
+
+        fp.seek(0)
+
+        for i, enquiry_row in enumerate(reader):
+            enquiry = enquiries[i]
+            for name in utils.ENQUIRY_OWN_FIELD_NAMES:
+                enquiry_val = getattr(enquiry, name)
+                self.assertEqual(
+                    enquiry_row[name], str(enquiry_val) if enquiry_val != None else ""
+                )
+            if enquiry.enquirer:
+                for enquirer_name in utils.ENQUIRER_FIELD_NAMES:
+                    # not sure why _meta.get_fields is returning "enquirer" but skip it
+                    if enquirer_name in ["enquirer"]:
+                        continue
+                    row_col = f"enquirer_{enquirer_name}"
+                    enquirer_val = getattr(enquiry.enquirer, enquirer_name)
+                    
+                    self.assertEqual(
+                        enquiry_row[row_col], str(enquirer_val) if enquirer_val != None else ""
+                    )
+            row_count += 1
+        self.assertEqual(row_count, 5)
--- a/app/enquiries/tests/test_views.py
+++ b/app/enquiries/tests/test_views.py
@@ -1,4 +1,5 @@
 import csv
+import io
 import pytest
 import pytz
 import random
@@ -22,6 +23,7 @@ from unittest import mock
 import app.enquiries.ref_data as ref_data
 import app.enquiries.views as enquiry_views
 import app.enquiries.tests.utils as test_utils
+
 from app.enquiries import utils
 from app.enquiries.models import Enquiry, Enquirer
 from app.enquiries.tests.factories import (
@@ -38,9 +40,6 @@ from app.enquiries.views import ImportEnquiriesView
 faker = Faker(["en_GB", "en_US", "ja_JP"])
 headers = {"HTTP_CONTENT_TYPE": "text/html", "HTTP_ACCEPT": "text/html"}
 headers_json = {"HTTP_CONTENT_TYPE": "text/html", "HTTP_ACCEPT": "application/json"}
-
-enquiry_field_names = [f.name for f in Enquiry._meta.fields]
-enquirer_field_names = [f.name for f in Enquirer._meta.fields]
 
 
 def canned_enquiry():
@@ -124,9 +123,7 @@ class EnquiryViewTestCase(test_utils.BaseEnquiryTestCase):
         omit_fields = ["owner"]
         ref_fields = ref_data.MAP_ENQUIRY_FIELD_TO_REF_DATA
 
-        for name in enquiry_field_names:
-            if name == "enquirer":
-                continue
+        for name in utils.ENQUIRY_OWN_FIELD_NAMES:
             factory_value = getattr(factory_item, name)
             db_value = response_item[name]
             if name in omit_fields:
@@ -151,6 +148,21 @@ class EnquiryViewTestCase(test_utils.BaseEnquiryTestCase):
 
             self.assertEqual(factory_value, db_value)
 
+    def assert_enquiry_equals_csv_row(self, enquiry, csv_row):
+        for name in utils.ENQUIRY_OWN_FIELD_NAMES:
+            enquiry_val = getattr(enquiry, name)
+            self.assertEqual(
+                csv_row[name], str(enquiry_val) if enquiry_val != None else ""
+            )
+        if enquiry.enquirer:
+            for enquirer_key in utils.ENQUIRER_FIELD_NAMES:
+                enquirer_val = getattr(enquiry.enquirer, enquirer_key)
+                csv_row_key = f"enquirer_{enquirer_key}"
+                self.assertEqual(
+                    csv_row[csv_row_key],
+                    str(enquirer_val) if enquirer_val != None else "",
+                )
+
     def create_enquiry_and_assert(self, enquiry):
         """Creates an Enquiry using the API and asserts on the response status"""
         response = self.client.post(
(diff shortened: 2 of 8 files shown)


app/settings/common.py
```diff
@@ -207,6 +207,8 @@
 CHAR_FIELD_MAX_LENGTH = 255
 ENQUIRIES_PER_PAGE = env.int('ENQUIRIES_PER_PAGE', default=10)
 IMPORT_TEMPLATE_FILENAME = 'rtt_enquiries_import_template.xlsx'
+EXPORT_OUTPUT_FILE_SLUG = 'rtt_enquiries_export'
```
saruniitr: I suggest adding date to the file name so we know when the data is exported.

app/settings/common.py
```diff
@@ -207,6 +207,8 @@
...
+EXPORT_OUTPUT_FILE_EXT = 'csv'
```
saruniitr: Not sure how useful it is in settings. Cchanging it to pdf, we get a file with pdf extension but it won't be valid.

app/enquiries/utils.py
```diff
@@ -92,3 +115,47 @@ def generate_import_template(file_obj):
             row = [str(v) for v in choice]
             current_sheet.append(row)
     book.save(file_obj)
+
+
+def enquiry_to_dict(e: models.Enquiry) -> dict:
```
saruniitr: This function and below function are not required, you can use `model_to_dict`

`from django.forms.models import model_to_dict`

app/enquiries/utils.py
```diff
@@ -1,10 +1,33 @@
...
 import app.enquiries.ref_data as ref_data
 from app.enquiries.models import Enquirer, Enquiry
 
 
+from django.db.models import QuerySet
+
+from app.enquiries import models, serializers
+
+
+ENQUIRER_FIELDS = models.Enquirer._meta.fields
```
saruniitr: All of this won't be required if we use `model_to_dict`

app/enquiries/tests/test_views.py
```diff
@@ -443,3 +450,28 @@ def test_import_template(self):
         self.assertIn(
             settings.IMPORT_TEMPLATE_FILENAME, response.get("Content-Disposition")
         )
+
+    def test_export_view(self):
+        enquiries = [EnquiryFactory() for i in range(5)]
```
saruniitr: EnquiryFactory.create_batch(5)

app/enquiries/tests/test_views.py
```diff
@@ -443,3 +450,28 @@ def test_import_template(self):
...
+        row_count = 0
+
+        self.assertTrue("text/csv" in response.get("Content-Type"))
+        self.assertIn(
+            f"{settings.EXPORT_OUTPUT_FILE_SLUG}.{settings.EXPORT_OUTPUT_FILE_EXT}",
+            response.get("Content-Disposition"),
+        )
+
+        for i, enquiry_row in enumerate(reader):
+            enquiry = enquiries[i]
```
saruniitr: you can use `mode_to_dict` here again then it is easy to access fields.
Also it doesn't seem like we are asserting `enquirer` fields.

app/enquiries/utils.py
```diff
@@ -92,3 +115,47 @@ def generate_import_template(file_obj):
...
+    :qs: Django queryset
+    :fp: file-like object. This includes Django HTTP response object
+    :return: void
+    """
+    output = []
+    writer = csv.DictWriter(fp, fieldnames=EXPORT_FIELD_NAMES)
+    writer.writeheader()
+
+    for e in qs.iterator():
+        data = enquiry_to_dict(e)
```
saruniitr: `created` and `modified` fields are exported as `datetime` isoformat() string. The microseconds may not be useful,
Instead of `2018-08-25 12:48:32.565000+00:00`, may be just `2018-08-25 12:48:32` is enough?

app/enquiries/tests/test_views.py
```diff
@@ -99,8 +102,10 @@ def assert_dicts_equal(self, expected, actual, exclude_keys=None):
                 continue
 
             self.assertEqual(value, expected[key])
-    
-    def assert_factory_enquiry_equals_enquiry_response(self, factory_item, response_item):
+
+    def assert_factory_enquiry_equals_enquiry_response(
+        self, factory_item, response_item
```
saruniitr: if you convert the factory_item to dict using `model_to_dict`, then it is easier to compare dictionaries. I think you can simply delete `created`, `modified` fields as they depend on when the test is executed.

app/enquiries/utils.py
kiddhustle: I did look at using `model_to_dict` but it doesn't handle relations (`foreign model`) how I would expect.

https://stackoverflow.com/questions/21925671/convert-django-model-object-to-dict-with-all-of-the-fields-intact#answer-29088221

app/enquiries/utils.py
kiddhustle: see above (i.e. `model_to_dict` doesn't handle relations well)

app/enquiries/tests/test_views.py
kiddhustle: amended

app/enquiries/tests/test_views.py
kiddhustle: see above (i.e. `model_to_dict` doesn't handle relations well)

Amended test to assert on the enquirer fields properties

app/enquiries/tests/test_views.py
kiddhustle: see above (i.e. `model_to_dict` doesn't handle relations well) and the keys are not the same

app/enquiries/utils.py
```diff
@@ -1,10 +1,33 @@
...
+
+from app.enquiries import models, serializers
+
+
+ENQUIRER_FIELDS = models.Enquirer._meta.fields
+ENQUIRY_FIELDS = models.Enquiry._meta.fields
+
+ENQUIRER_FIELD_NAMES = [f.name for f in ENQUIRER_FIELDS]
+ENQUIRY_FIELD_NAMES = [f.name for f in ENQUIRY_FIELDS]
+ENQUIRY_FIELD_NAMES.remove("enquirer")
```
elcct: That will make this variable misleading, as `enquirer` is a also an Enquiry field.

app/enquiries/tests/test_utils.py
```diff
@@ -1,4 +1,13 @@
+import codecs
+import csv
+import io
+
+from app.enquiries import models, utils
 import app.enquiries.tests.utils as test_utils
+from app.enquiries.tests.factories import EnquiryFactory
+
+enquiry_field_names = [f.name for f in models.Enquiry._meta.fields]
```
elcct: You seem to have these in `app/enquiries/utils.py`.

app/enquiries/utils.py
```diff
@@ -1,10 +1,33 @@
...
+
+EXPORT_FIELD_NAMES = ENQUIRY_FIELD_NAMES + [
+    "enquirer_" + n for n in ENQUIRER_FIELD_NAMES
+]
+
+foreign_keys = ("owner", "enquirer")
```
elcct: This variable sounds too generics. It is also weird that you hardcode foreign keys, but get field names from the model object. I think it should be possible to get a list of foreign keys from the model.

app/enquiries/utils.py
```diff
@@ -92,3 +115,47 @@ def generate_import_template(file_obj):
...
+    # return serializers.EnquiryDetailSerializer(e).data
+    output = {}
+
+    for f in ENQUIRY_FIELDS:
+        if f.name not in foreign_keys:
+            output[f.name] = getattr(e, f.name)
+        else:
+            if f.name == "enquirer":
+                output.update(enquirer_to_dict(getattr(e, f.name)))
+            else:
```
elcct: This loop looks too complicated. You could probably split it into smaller loops.

app/enquiries/tests/test_views.py
```diff
@@ -137,6 +142,22 @@ def assert_factory_enquiry_equals_enquiry_response(self, factory_item, response_
 
             self.assertEqual(factory_value, db_value)
 
+    def assert_enquiry_equals_csv_row(self, enquiry, csv_row):
+        for name in enquiry_field_names:
+            if name == "enquirer":
+                for enquirer_key in enquirer_field_names:
```
elcct: You could probably check the `enquirer` outside the loop. Otherwise this looks too complex and you are checking for `name` in every iteration.

app/enquiries/tests/test_utils.py
kiddhustle: yes I think I had to add these twice because of the issue with delayed PRs can store them in a single place

package.json
```diff
@@ -1,30 +0,0 @@
-{
```
reupen: Was removing this intended?

app/enquiries/utils.py
reupen: `.fields` is an internal API so we should avoid using that: 
https://github.com/django/django/blob/00ff7a44dee91be59a64559cadeeba0f7386fd6f/django/db/models/options.py#L439-L441

`.get_fields()` is the public one: https://docs.djangoproject.com/en/3.0/ref/models/meta/#retrieving-all-field-instances-of-a-model

There are these attributes available on fields for further filtering:
https://docs.djangoproject.com/en/3.0/ref/models/fields/#attributes-for-fields-with-relations

app/settings/common.py
kiddhustle: data has been added

app/enquiries/utils.py
kiddhustle: amended to use `get_fields()`

app/enquiries/tests/test_utils.py
```diff
@@ -62,3 +69,30 @@ def test_helper_logout(self):
...
+        utils.export_to_csv(qs, fp)
+        reader = csv.DictReader(fp)
+        row_count = 0
+
+        fp.seek(0)
+
+        for i, enquiry_row in enumerate(reader):
+            enquiry = enquiries[i]
+            for name in utils.ENQUIRY_FIELD_NAMES:
+                print('name', name)
```
mcbhenwood: Debug code?

app/enquiries/tests/test_utils.py
```diff
@@ -62,3 +69,30 @@ def test_helper_logout(self):
...
+                if name == "enquirer":
```
mcbhenwood: Why do we skip "enquirer"?

app/enquiries/tests/test_utils.py
reupen: If these are needed it would be good to update them to use `.get_fields()` as well.

app/enquiries/tests/test_utils.py
kiddhustle: removed

app/enquiries/tests/test_utils.py
kiddhustle: added tests for enquirer fields

package.json
kiddhustle: must have been a merge mistake. re-added

app/enquiries/tests/test_views.py
kiddhustle: refactored this

app/enquiries/utils.py
kiddhustle: refactored

app/enquiries/utils.py
kiddhustle: refactored

app/enquiries/tests/test_utils.py
kiddhustle: amended

app/enquiries/utils.py
kiddhustle: created new variable: `ENQUIRY_OWN_FIELD_NAMES`

kiddhustle:
@reupen @elcct @mcbhenwood have made amendments re the feedback

app/enquiries/tests/test_utils.py
elcct: This still appears to be using `.fields`?

kiddhustle:
@elcct  @reupen  thanks for the review. Redundant code has been removed will now merge

PR #33, opened 2020-03-30
Refactor CSS into separate files using BEM naming

hnryjmes:
HTML and CSS changes including:
- CSS now in separate files according to specificity
- CSS selectors follow BEM (Block Element Modifier) convention


app/enquiries/templates/base.html
```diff
@@ -18,7 +18,11 @@
   <link rel="apple-touch-icon" href="{% static 'assets/images/govuk-apple-touch-icon.png' %}">
 
   <!--[if !IE 8]><!-->
-  <link href="{% static 'css/application.css' %}" rel="stylesheet">
+  <link href="{% static 'css/enquiry_delete.css' %}" rel="stylesheet">
```
ian-leggett: You shouldn't need separate CSS files here, I would bundle the dist version as `application.css`

frontend/sass/enquiries.scss
```diff
@@ -0,0 +1,54 @@
...
+
+.govuk-header__content {
+    padding-left: 0px;
+}
+
+.govuk-label {
+    font-weight: bold;
+}
+
+ol {
```
ian-leggett: If this is a global rule then maybe this needs to go into a `settings.scss` look at https://github.com/alphagov/govuk-frontend/tree/master/src/govuk/settings

frontend/sass/enquiries.scss
```diff
@@ -0,0 +1,54 @@
...
+.govuk-header__logo, .govuk-header__content {
+    display: inline-block;
+}
+
+.govuk-header__container {
+    border-bottom: 10px solid white;
+}
+
+.govuk-header__link--service-name {
+    font-size: 30px;
```
ian-leggett: Is this an official Gov font size?

frontend/sass/enquiry_delete.scss
```diff
@@ -0,0 +1,15 @@
+$govuk-assets-path: '/static/assets/';
+
+@import "node_modules/govuk-frontend/govuk/all";
```
ian-leggett: We don't need to import this here, this should be in the root SCSS file such as `application.scss`

frontend/sass/enquiry_detail.scss
```diff
@@ -0,0 +1,16 @@
+$govuk-assets-path: '/static/assets/';
+
+@import "node_modules/govuk-frontend/govuk/all";
```
ian-leggett: See above comment

frontend/sass/enquiry_edit.scss
```diff
@@ -0,0 +1,4 @@
+$govuk-assets-path: '/static/assets/';
+
+@import "node_modules/govuk-frontend/govuk/all";
```
ian-leggett: See above comment

frontend/sass/enquiry_list.scss
```diff
@@ -0,0 +1,95 @@
+$govuk-assets-path: '/static/assets/';
+
+@import "node_modules/govuk-frontend/govuk/all";
```
ian-leggett: See above comment

saruniitr:
Thanks for your reviews, received two approvals, merging it

PR #36, opened 2020-03-31
Fix styling issues

saruniitr:
Filters branch modified templates which introduced these issues.

Base template modified to include header and footer so this removed Edit button from detail view.
Edit view was having additional footer at the top.
Pagination links were also missing from the list view.


hnryjmes:
I won't be able to run this locally so I can't review this properly - that said I'd agree with any changes that restore the earlier state, because I believe the back button and edit button are currently designed to be placed outside of the main content and there doesn't seem much point in redesigning at this late stage.

saruniitr:
Thanks for the review, merging with internal review as these are only template refactoring changes. No functional changes are done in this PR.

PR #79, opened 2020-04-05
Fixes for CSV file import of Enquiries

saruniitr:
## Change description

The feature to import enquiries using CSV file is merged but we found issues during initial testing. This PR fixes these issues.

- CSV files created on Windows system were not able to import because of different encoding, backend logic updated to use windows-1252 decoding in this case.
- Refactored and simplified the code that creates instances using CSV rows.
- Added missing links in the UI to import/export enquiries
- Added additional error checks to detect empty file or file with only header etc
- Multiple import templates were created, removed the duplicated template


app/enquiries/views.py
```diff
@@ -206,38 +204,6 @@ def form_invalid(self, form):
         return response
 
 
-class EnquiryAdd(APIView):
-    renderer_classes = [TemplateHTMLRenderer]
```
saruniitr, line 210: This was the duplicated view which is removed and also removed the associated tests above.

app/enquiries/tests/test_views.py
```diff
@@ -403,27 +403,6 @@ def test_enquiry_import_rendered(self):
         self.assertNotContains(response, "File import successfully completed.")
         self.assertEqual(response.status_code, status.HTTP_200_OK)
 
-    def test_enquiry_import_rendered_with_error(self):
-        url = reverse("enquiries-import") + "?errors"
```
saruniitr, line 407: these tests no longer required as belong to the duplicated view (`EnquiryAdd`) that has been removed (please see further below)

app/enquiries/views.py
```diff
@@ -264,27 +231,41 @@ class ImportEnquiriesView(TemplateView):
...
+            buf = BytesIO()
+            for c in f.chunks(chunk_size=settings.UPLOAD_CHUNK_SIZE):
+                buf.write(c)
+
+            # Since the processing happens on PaaS we cannot determine User OS
+            # so assume utf-8 by default and switch to windows encoding in case of error
+            try:
+                buf.seek(0)
+                decoded = buf.read().decode("utf-8")
+            except Exception:
```
elcct, line 262: I wonder if it would be better to use `chardet` to detect the incoming file's encoding, as this way looks a bit "hacky". You can see an example of this at work here https://github.com/uktrade/data-hub-api/blob/develop/datahub/core/admin_csv_import.py#L5

saruniitr: I don't consider it as hacky as `utf-8` should cover most of the cases. The other type is windows which is also being considered. However using chardet seems better so modified it.
There is confidence level attached for each result so it may fail under edge cases but should cover all common cases.

app/enquiries/templates/enquiry_import.html
```diff
@@ -35,22 +30,23 @@ <h3 class="govuk-heading-s">
...
     </h3>
     <ol class="govuk-list govuk-list--number">
-        <li>Click <strong>Download template </strong> to download the CSV template.</li>
-        <li>Open the CSV template and insert the enquiry data.</li>
-        <li>Save the CSV file.</li>
-        <li>Click <strong>Choose file</strong>, locate your CSV file and then click <strong>Open</strong>.</li>
-        <li>Click Upload file.</li>
+        <li>Click <a href="{% url 'import-template' %}"><strong>Download template </strong></a> to download the CSV template</li>
+        <li>Open the template and insert the enquiries data in the 'enquiries' sheet</li>
+        <li>Save the sheet as CSV file (eg Enquiries_DDMMYYYY.csv)</li>
```
reupen, line 35: I'd suggest also adding an instruction to save as `CSV UTF-8` when using Excel (as just `CSV` will save in the user's code page which can still behave unpredictably):

csv utf-8

saruniitr: ok, will add instruction as suggested

saruniitr: Added below to the instructions,
`Select CSV utf-8 or windows-1252 from the available encoding options`

app/enquiries/views.py
```diff
@@ -264,59 +232,89 @@ class ImportEnquiriesView(TemplateView):
...
 
+        file_obj = request.FILES.get(enquiries_key)
+        if not (
+            file_obj
+            and file_obj.name.endswith(".csv")
+            and file_obj.content_type in settings.IMPORT_ENQUIRIES_MIME_TYPES
+        ):
+            messages.error(
+                self.request,
+                f"Input file is not a CSV file, detected type: {file_obj.content_type}",
```
mcbhenwood, line 292: It is possible that if a user uploads a properly formatted CSV file but omits the ".csv" file extension then this message will read to the user: "Input file is not a CSV file, detected type: text/csv" which I think will be confusing for them.

app/enquiries/tests/test_views.py
```diff
@@ -446,25 +425,31 @@ def test_enquiry_import_post_success(self):
...
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertEqual(final_count, num_items)
+            response = self.client.post(
+                reverse("import-enquiries"), {"enquiries": upload}, follow=True
+            )
+
+            db_enquiries = Enquiry.objects.all()
+            final_count = db_enquiries.count()
+            self.assertNotContains(response, ImportEnquiriesView.ERROR_HEADER)
+            self.assertContains(response, f"Successfully posted {num_items} records!")
```
mcbhenwood, line 439: I appreciate it wasn't introduced by this PR but the exclamation mark can probably go. It's a bit unprofessional IMO.

app/enquiries/templates/enquiry_import.html
```diff
@@ -4,21 +4,16 @@
...
+    <h1 class="govuk-heading-l">Import Enquiries</h1>
     {% if messages %}
     <div class="govuk-error-summary {% if has_success%}govuk-error-summary--success{% endif %}" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary">
         <h2 class="govuk-error-summary__title" id="error-summary-title">
-            {% if has_errors %}
-                Error - File import has failed:
-            {%  else %}
-                File import successfully completed.
-            {% endif %}
+            There is a problem
```
mcbhenwood, line 16: Super-picky but I feel this needs a full stop.

saruniitr: will fix it

app/enquiries/tests/test_views.py
saruniitr: agree, will update this.

app/enquiries/views.py
saruniitr: I think content type is decided based on the extension. I did the same test and got the error message shown below so we are good,
`Input file is not a CSV file, detected type: application/octet-stream`

saruniitr:
Thanks for your reviews, received two approvals, merging it.

Issue #57, opened 2020-04-01
Reference data value changes (Enquiry quality)


rahulbist:
IST have requested to make the following changes on the ref data for the dropdown field “Enquiry quality”:

Modify “FDI or likely FDI” to “FDI”

saruniitr:
Created PR https://github.com/uktrade/enquiry-mgmt-tool/pull/87 to fix this issue.

PR #87, opened 2020-04-07
Update Enquiry quality field value in reference data

saruniitr:
## Change description

Updates one of the Enquiry quality field name based on the request from Service team.

Fixes the issue https://github.com/uktrade/enquiry-mgmt-tool/issues/57

Updated reference data and generated migrations.


app/enquiries/ref_data.py
```diff
@@ -40,7 +40,7 @@ class Quality(models.TextChoices):
     NON_FDI = "NON_FDI", _("Non-FDI")
     POTENTIALLY_NON_FDI = "POTENTIALLY_NON_FDI", _("Potentially Non-FDI")
     POTENTIALLY_FDI = "POTENTIALLY_FDI", _("Potentially FDI")
-    LIKELY_FDI = "LIKELY_FDI", _("FDI or likely FDI")
+    FDI = "FDI", _("FDI")
```
currycoder, line 43: Do we have data in production yet?  I believe that we need to have a data migration to move existing "LIKELY_FDI" values to "FDI"

Issue #46, opened 2020-04-01
Paginating with filter conditions


rahulbist:
When applying single or multiple filter conditions within the Enquiry Management tool on the homepage (enquiries screen), when paginating through, the applied filter is erase, resulting in all enquiries being displayed

rahulbist:
This bug now only exists when paginating via page numbers. If paginating through the "Next" or "Previous" button, the filter condition is kept/remains

PR #89, opened 2020-04-14
Fix pagination in combination with filters

saruniitr:
## Change description

Both pagination and filters use query parameters to provide variables
to backend. They both individually work ok but the parameters gets
overwritten if used in combination. Eg apply some filters and select
a particular page from the list of results. In this case once a page
is selected then existing query parameters gets overwritten with the
selected page and because of this filters also gets cleared.

The fix is to append the variables instead of overwriting them which
is implemented using custom template tags.

This PR fixes https://github.com/uktrade/enquiry-mgmt-tool/issues/46

### DIFF
--- a/app/enquiries/templatetags/enquiries_extras.py
+++ b/app/enquiries/templatetags/enquiries_extras.py
@@ -91,3 +91,31 @@ def query_params_has_value(value, param_key, query_params):
 @register.simple_tag
 def query_params_value_selected(value, param_key, query_params, text='selected'):
     return f' {text}' if str(value) in query_params.getlist(param_key) else ''
+
+
+@register.simple_tag
+def query_params_with_pagination(query_params, page):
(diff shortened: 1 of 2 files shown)


app/enquiries/templatetags/enquiries_extras.py
```diff
@@ -91,3 +91,31 @@ def query_params_has_value(value, param_key, query_params):
 @register.simple_tag
 def query_params_value_selected(value, param_key, query_params, text='selected'):
     return f' {text}' if str(value) in query_params.getlist(param_key) else ''
+
+
+@register.simple_tag
```
currycoder, line 96: I think these tag functions need some tests.

mforner13:
Closing this PR as another PR has been merged to fix this issue in another way (#103).

mforner13:
Closing this PR as another PR has been merged to fix this issue in another way (#103).

Issue #80, opened 2020-04-06
Company name filter not displaying full string term


rahulbist:
When applying multiple string terms to "Company name", the UI searches the entire string inputted however, only displays the first string. For example:

If i search the string "Automated test - company name" on the company search filter, the entire string is searched but on the filter only "Automated" is displayed:

![image](https://user-images.githubusercontent.com/61739458/78541364-6beefb80-77ed-11ea-8c76-d2f9a9c7fac1.png)

PR #90, opened 2020-04-14
Fix company name filter not showing full search term

saruniitr:
## Change description

If the company name search is performed with multiple words separated
by space then upon fetching the results only the first word is shown
in the search field. The results themselves are correct except that
only first word is shown in the search field.

Eg if we search for `testing company` then after fetching results the search term is shown as `testing`

This is just due to not enclosing input field value in double quotes.

This PR fixes https://github.com/uktrade/enquiry-mgmt-tool/issues/80


rafenden:
Would be nice to have some related tests.

PR #91, opened 2020-04-21
Pagination tests

ian-leggett:
## Description
This adds tests around the pagination.


app/enquiries/e2e_tests/specs/pagination.test.js
```diff
@@ -0,0 +1,35 @@
+const { pagination, results } = require('../selectors')
+
+describe('Pagination', () => {
+  before(() => {
+    cy.login('/enquiries/')
+  })
+
+  beforeEach(()=>{
+    Cypress.Cookies.preserveOnce('sessionid')
```
PippoRaimondi, line 9: could you have this in the `before` rather than the `beforeEach`?

app/enquiries/e2e_tests/specs/pagination.test.js
```diff
@@ -0,0 +1,35 @@
...
+    cy.get(pagination.pageLink(1)).should('have.text', '1')
+    cy.get(pagination.lastPageLink).should('have.text', 'Next')
+    cy.get(pagination.lastPageLink).click()
+    cy.get(pagination.headerPageNumber).should('have.text', 'Page 2 of 4')
+    cy.get(pagination.pageLink(1)).should('have.text', 'Previous')
+    cy.get(pagination.pageLink(1)).click()
+    cy.get(pagination.pageLink(1)).should('have.text', '1')
+  })
+
+  it('should update the enquiries', () => {
```
PippoRaimondi, line 30: tempted in having these assertions imbedded with the tests above... so every time we navigate pages, assert the enquiries changed as well, that would make this test redundant and make the above 2 more meaningful!

setup-fixtures.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -xe
+
+python manage.py loaddata app/enquiries/fixtures/users.json
+python manage.py loaddata app/enquiries/fixtures/enquiries.json
+python manage.py createsuperuser --no-input
```
PippoRaimondi, line 5: just realised the `start.sh` has the creation of the superuser, so we could possibly remove it from here?

cypress.json
```diff
@@ -1,5 +1,5 @@
 {
-  "integrationFolder": "app/enquiries/e2e_tests",
+  "integrationFolder": "app/enquiries/e2e_tests/specs/",
```
rafenden, line 2: Perhaps, it's worth moving it to the `cypress` dir in the root level?

app/enquiries/e2e_tests/specs/pagination.test.js
ian-leggett: Perfect I thought I needed `beforeEach` as the cookie would get wiped out on `before` this is much better 👍

app/enquiries/e2e_tests/specs/pagination.test.js
ian-leggett: Makes sense thanks

setup-fixtures.sh
ian-leggett: Removed

cypress.json
ian-leggett: Makes sense thanks, just moved it 👍

PR #95, opened 2020-04-30
Add and refactor e2e tests around edit enquiry

ian-leggett:
## Description
This is the final E2E tests around the edit enquiry journey. I have refactored previous work to be more DRY and added the final bits of code to finish the testing.

### DIFF
--- a/cypress/e2e_tests/support/commands.js
+++ b/cypress/e2e_tests/support/commands.js
@@ -1,18 +1,20 @@
-Cypress.Commands.add('findDetailsSection', number => {
-  return number === 0
-    ? cy.get('main').find('div > div').eq(number).find('h2')
-    : cy.get('main').find('div > div').eq(number).find('h3')
-})
+Cypress.Commands.add('findDetailsSection', number =>
+  cy
+    .get('main')
+    .find('div > div')
+    .eq(number)
+    .find(number === 0 ? 'h2' : 'h3')
+)
 
 const populateField = (type, name, value) => {
-  const fieldTypes = {
-    select: () => cy.get(`select[name=${name}]`).select(value),
-    textarea: () => cy.get(`textarea[name=${name}]`).clear().type(value),
-    default: () => cy.get(`input[name=${name}]`).clear().type(value),
+  switch (type) {
+    case 'select':
+      return cy.get(`select[name=${name}]`).select(value)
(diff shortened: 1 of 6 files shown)


cypress/e2e_tests/support/commands.js
```diff
@@ -1 +1,22 @@
-Cypress.Commands.add('findDetailsSection', number => cy.get('main').find('div > div').eq(number).find('h3'))
+Cypress.Commands.add('findDetailsSection', number => {
+  return number === 0
+    ? cy.get('main').find('div > div').eq(number).find('h2')
+    : cy.get('main').find('div > div').eq(number).find('h3')
+})
```
peterhudec, line 5: You could move the condition to the last `.find()`:
```js
Cypress.Commands.add('findDetailsSection', number =>
  cy.get('main')
    .find('div > div')
    .eq(number)
    .find(number === 0 ? 'h2' : 'h3')
)
```

cypress/e2e_tests/support/commands.js
```diff
@@ -1 +1,22 @@
...
+const populateField = (type, name, value) => {
+  const fieldTypes = {
+    select: () => cy.get(`select[name=${name}]`).select(value),
+    textarea: () => cy.get(`textarea[name=${name}]`).clear().type(value),
+    default: () => cy.get(`input[name=${name}]`).clear().type(value),
+  }
+  return typeof fieldTypes[type] !== 'function'
+    ? fieldTypes['default']()
+    : fieldTypes[type]()
+}
```
peterhudec, line 16: I also tend to use objects for this kind of logic, just to realize that a switch statement is probably more readable:
```js
switch (type) {
  case 'select':
    return cy.get(`select[name=${name}]`).select(value)
  case 'textarea':
    return cy.get(`textarea[name=${name}]`).clear().type(value)
  default:
    return cy.get(`input[name=${name}]`).clear().type(value)
}
```

cypress/e2e_tests/support/commands.js
ian-leggett: Thanks much better 👍

cypress/e2e_tests/support/commands.js
ian-leggett: I agree, in this example its much clearer as a switch statement

PR #97, opened 2020-05-04
Add end to end test for filtering

peterhudec:


cypress/e2e_tests/specs/filter.test.js
```diff
@@ -0,0 +1,558 @@
...
+    [pk]: `${first_name} ${last_name}`,
+  }),
+  {},
+)
+
+// The result list items only show a subset of the enquiry data and we need
+// to make assertions about the information not shown. The constants
+// ENQUIRIES and ENQUIRERS constants are maps of primary keys to enquiries and
+// enquirers fixtures respectively, which we are using to make assertions.
+const fixtureReducerc = (a, {pk, fields}) => ({ ...a, [pk]: fields })
```
rafenden, line 16: Typo at the end.
```suggestion
const fixtureReducer = (a, {pk, fields}) => ({ ...a, [pk]: fields })
```

cypress/e2e_tests/specs/filter.test.js
```diff
@@ -0,0 +1,558 @@
...
+  added: 'Added to Data Hub',
+  nonFdi: 'Non-FDI',
+  nonResponsive: 'Non-responsive',
+  awaiting: 'Awaiting response from Investor',
+  postProgressing: 'Post progressing',
+  sent: 'Sent to Post',
+  engaged: 'Engaged in dialogue',
+}
+
+const filterLabels = filters =>
```
ian-leggett, line 35: Is it worth moving this to `commands.js`

cypress/e2e_tests/specs/filter.test.js
```diff
@@ -0,0 +1,558 @@
...
+  Object.entries(filters)
+    .map(([k, v]) =>
+      typeof v === 'string'
+        ? `${k}: ${v}`
+        : k
+     )
+    .join(', ')
+
+const getInputByLabel = (label) =>
```
ian-leggett, line 44: Is it worth moving this to `commands.js`

cypress/e2e_tests/specs/filter.test.js
```diff
@@ -0,0 +1,558 @@
...
+  cy.contains('label', label)
+    .parent()
+    .find('input')
+
+const setFilters = filters =>
```
ian-leggett, line 49: Is it worth moving this to `commands.js`

cypress/e2e_tests/specs/filter.test.js
```diff
@@ -0,0 +1,558 @@
...
+        // Date input doesn't allow empty string to be typed in
+        ? input.clear()
+        : input.type(val)
+      : val
+        ? input.check()
+        : input.uncheck()
+
+  })
+
+const assertResults = (assert, expectedTotal) => {
```
ian-leggett, line 63: Is it worth moving this to `assertions.js`

cypress/e2e_tests/specs/filter.test.js
```diff
@@ -0,0 +1,558 @@
...
+    .contains(`${expectedTotal} enquiries`)
+
+  expectedTotal && cy.get('article')
+    .find('ol li')
+    .as('items')
+    .should('have.length', expectedTotal > 10 ? 10 : expectedTotal)
+    .each(assert)
+}
+
+const testFilters = ({ filters, assertItem, expectedTotal }) => {
```
ian-leggett, line 74: Is it worth moving this to `assertions.js`

peterhudec: The thing is that this function and all the other ones you mention in the above comments only make sense in this spec file and I would only move generic stuff to `./support`. A possible candidate would be the `getInputByLabel`, which could be very useful, but unfortunately, it is not trivial to make it work with all possible configuration of inputs and labels. The `getInputByLabel` here assumes a certain DOM structure of the filters so it's not really reusable.

PR #98, opened 2020-05-07
Allow for db fixture reset in e2e tests

mcbhenwood:
Add a URL endpoint which resets the database to its initial fixture state. This is to assist e2e tests.

Naturally the URL is not exposed under normal circumstances.


mcbhenwood:
Closing, going to rebase

PR #99, opened 2020-05-07
Allow for db fixture reset in e2e tests

mcbhenwood:
Add a URL endpoint which resets the database to its initial fixture state. This is to assist e2e tests.

Naturally the URL is not exposed under normal circumstances.

Note that CircleCI build is currently broken for Cypress test for many reasons. We are fixing the build and so stripping back the CI job to the unit test level and rebuilding from there.

Future PRs will reinstate the Cypress e2e tests and maintain a green build.


mcbhenwood:
Clsoing due to rebase (again)

PR #101, opened 2020-05-10
Enable "resetting" of db state for e2e tests

mcbhenwood:
This PR introduces an API endpoint which allows for an e2e testing client to reset the database to a known state so that each e2e test case can be truly independent.

Yay super awesome etc BUT of course this kind of thing is not _so_ yay super awesome etc from a security point of view. Therefore also in this PR are the settings and restrictions which prohibit this facility from being exposed by accident or indeed at all within a production environment. I discussed this approach with the Architects Community a few weeks back and it was approved. [ping @munhitsu ]

The idea for this method is that it's called at the start of every test to ensure test cleanliness. While it's great to have the fixtures reset before every single test, that also wipes all the user records each time and so the e2e tests will be unable to access the app because they don't have a valid user session.

Getting a user session is hard because this app uses Staff SSO and at time of writing, hooking up a test / CI instance of an app to a controllable Staff SSO instance is non-trivial. So the answer commonly has been to set up a Django superuser, expose the admin login form and go from there. Personally I think that performing user testing with a superuser / admin account is a bad practice. So I'd like to avoid this.

For this reason, the ‘reset fixtures’ method also creates for you a _standard_ user of your choosing, logs them in and returns their new Session ID in the cookies of the response header. See the body of the PR for how this is specified and set up.

There are two other points about this PR which need to be noted:

  1. It removes two settings modules which were totally unused

  2. The master branch of this repo has been broken (in CI terms) for a while now. The Cypress tests have been failing. Seeing as _THIS_ PR is intended to fix the e2e testing pain and work towards a green build, it’s important that we start from a known green state. Therefore this PR also disables Cypress testing in the CI environment to achieve a green build for a codebase with these changes, then we can reinstate the Cypress suite piece by piece on top of this arriving finally at a totally clean build.


mcbhenwood:
N.B. There are a couple of changes coming down the track for this regarding docker compose setup needing to be different between unit testing and CI testing.

app/settings/common.py
```diff
@@ -205,6 +205,15 @@
...
 STATIC_ROOT = os.path.join(APP_ROOT, 'enquiries', 'static')
 
+# This setting alone will NOT enable the "Test Fixture API" facility -
+# By default the URL to it is not enabled. See README.md section
+# "Allowing for Fixture Reset during e2e tests" for details how to
+# enable the URL and turn on the Test Fixture API.
+if env('ALLOW_TEST_FIXTURE_SETUP', default=None) == 'allow':
+    ALLOW_TEST_FIXTURE_SETUP = True
+else:
+    ALLOW_TEST_FIXTURE_SETUP = False
```
peterhudec, line 215: This can be slightly simplified:
```python
ALLOW_TEST_FIXTURE_SETUP = env('ALLOW_TEST_FIXTURE_SETUP', default=None) == 'allow'
```

app/testfixtureapi/tests.py
```diff
@@ -0,0 +1,190 @@
...
+    settings.ALLOW_TEST_FIXTURE_SETUP = True
+    Client().post(
+        RESET_URL,
+        SEED_USER_DATA,
+        content_type='application/json'
+    )
+    created_user = Owner.objects.get(email=SEED_USER_DATA['email'])
+    expected_value = SEED_USER_DATA[seed_user_property]
+    actual_value = getattr(created_user, seed_user_property)
+    assert actual_value == expected_value
```
peterhudec, line 151: Wouldn't it be simpler to assert the values of all three properties instead of parametrizing the test? It would also ensure the coherence of the three values.
```python
@pytest.mark.django_db
def test_seed_user_created_as_specified(seed_user_property, settings):
    settings.ALLOW_TEST_FIXTURE_SETUP = True
    Client().post(
        RESET_URL,
        SEED_USER_DATA,
        content_type='application/json'
    )
    created_user = Owner.objects.get(email=SEED_USER_DATA['email'])
    for attr in ['username', 'first_name', 'last_name']:
        expected_value = SEED_USER_DATA[attr]
        actual_value = getattr(created_user, attr)
        assert actual_value == expected_value
    # or just
    assert created_user.username == SEED_USER_DATA['username']
    assert created_user.first_name == SEED_USER_DATA['first_name']
    assert created_user.last_name == SEED_USER_DATA['last_name']
```

app/testfixtureapi/views.py
```diff
@@ -0,0 +1,84 @@
...
+
+    The database will have its Enquiry, Enquirer and Owner objects removed and
+    reset to the state in the fixtures files. In addition a 'seed user' will be
+    created according to the data supplied in your payload and that user will
+    be automatically logged in (to simplify the e2e testing cycle).
+    """
+    authentication_classes = (CsrfExemptSessionAuthentication,)
+
+    def post(self, request, *args, **kwargs):
+        if settings.ALLOW_TEST_FIXTURE_SETUP != True:
```
peterhudec, line 57: It woud be more idiomatic with the `not` keyword:
```python
if not settings.ALLOW_TEST_FIXTURE_SETUP:
    return Response(status=status.HTTP_404_NOT_FOUND)
```

PR #104, opened 2020-06-02
Add mock-sso to docker-compose

peterhudec:


sample_env
```diff
@@ -33,11 +33,24 @@ CELERY_TIMEZONE=Europe/london
 # Staff SSO/OAuth2 settings
 FEATURE_ENFORCE_STAFF_SSO_ENABLED=1
 
-AUTHBROKER_URL=contact-web-ops-for-details
+# The AUTHBROKER_URL needs to be accessible from the browser for the
+# authorization redirect and also from the Docker container for the access token
+# POST request, and the only way to do this is to use the
+# docker.for.mac.localhost host provided by Docker.
+AUTHBROKER_URL=http://docker.for.mac.localhost:8080
```
mcbhenwood, line 40: Is `docker.for.mac.localhost` a standard host entry in Docker-enabled Macs? It doesn't resolve on mine (v 2.2.0.5 of Docker Desktop)

mcbhenwood: Of course, the _real_ fix here would be to alter django-staff-sso-client to accept separate URL for auth redirect, token, introspection etc. and then to pass them here (then the 'browser viewpoint' redirect-to-auth URL could be at `localhost` whereas the introspection etc inside Dover containers could be at `mock-sso`)

I've added a ticket to the tech backlog to do this as a separate task: https://trello.com/c/0Ck8Z7b3/113-fix-django-staff-sso-client-lib

peterhudec: I was trying hard to find a mention of `docker.for.mac.localhost` in Docker docs, but it's only mentioned in their changelog. The docs mention the `host.docker.internal`, which according to the changelog is a more robust alternative for `docker.for.mac.localhost`, but that was not working on my Mac. The readme mentions both though.

rafenden: Why not to use `localhost:8080`?

CircleCI tests will fail as `docker.for.mac.localhost` will not be resolved on Linux hosts.

mcbhenwood: Hi @rafenden `localhost` will not work for the backend of the process. In the Django container that will resolve to the Django container, not the mock-sso one.

Currently the CI tests aren't using the mock-sso authentication route so it doesn't matter. (The e2e tests use a back-door setup API I wrote for the tool instead)

PR #106, opened 2020-06-09
Convert the CRM input field to an autocomplete

peterhudec:
Replaces the CRM input field in the _edit enquiry_ form with a restyled [`django-autocomplete-light`](https://django-autocomplete-light.readthedocs.io/en/master/) widget. Adds a new `/dh-adviser-search` endpoint for the widget to fetch values. The widget's value will be the selected adviser's email on form submission.


app/enquiries/forms.py
```diff
@@ -18,3 +18,4 @@ class EnquiryForm(ModelForm):
     class Meta:
         model = Enquiry
         fields = "__all__"
+        widgets = {'crm': autocomplete.Select2(url="/dh-adviser-search")}
```
mcbhenwood, line 21: We ought either put the url in a setting, _or_ user `reverse()` to derive it here. (Probably the latter)

app/enquiries/templates/enquiry_edit.html
```diff
@@ -283,3 +287,4 @@ <h3 class="govuk-heading-m">Data Hub</h3>
 </div>
 
 {% include 'partials/govuk_footer_partial.html' %}
+{{ form.media }}
```
mcbhenwood, line 290: Optional: Is this OK in the body? Does it not belong in the `<head>` section of the page?

sample_env
```diff
@@ -11,6 +11,7 @@ DJANGO_SENTRY_DSN=sentry-dsn
 ENQUIRIES_PER_PAGE=10
 
 # Data Hub variables
+DATA_HUB_API_URL=http://dh-api-mock:8001
 DATA_HUB_METADATA_URL=http://dh-api-mock:8001/metadata
 DATA_HUB_COMPANY_SEARCH_URL=http://dh-api-mock:8001/company/search
 DATA_HUB_CONTACT_SEARCH_URL=http://dh-api-mock:8001/contact/search
```
mcbhenwood, line 17: I see the pain, but.... clearly the previous dev(s) put each separate DH API endpoint into a separate env var. So I think in your solution you need either to march in step and replace this with a `DATA_HUB_ADVISER_SEARCH_URL` which contains the `/adviser` path, _OR_ to keep your `DATA_HUB_API_URL` env var and get rid of their `DATA_HUB_API_URL` etc.

app/enquiries/templates/enquiry_edit.html
peterhudec: They have it in the body in [their docs](https://django-autocomplete-light.readthedocs.io/en/master/tutorial.html#using-autocompletes-outside-the-admin), but it is now quite normal to have the scripts the last thing in the body to ensure that the whole page is loaded when the script kicks in.

sample_env
peterhudec: Yeah, my plan was to get rid of all the other ones, kind of don't feel like doing it and I'll just do it as they did.

app/enquiries/forms.py
```diff
@@ -12,9 +13,45 @@ class Meta:
         fields = "__all__"
 
 
+class AutocompleteField(ChoiceField):
+    widget = autocomplete.Select2(url=reverse_lazy('dh-adviser-search'))
+
+    # We need to disable deep copying so, that our
+    # This is the sole reason why we subclass ChoiceField
+    # The sole reason why we need to subclass this class is that we need to
+    # disable deep copying, which prevents saving the new value to the field.
```
mcbhenwood, line 22: Optional: Comment seems to be muddled.

cypress/e2e_tests/specs/pokus.test.js
```diff
@@ -0,0 +1,43 @@
...
+    cy.server()
+    cy.route({
+      method: 'GET',
+      url: '/dh-adviser-search',
+      response: {
+        results: [
+          {text: 'Andy Pipkin', id: 'Andy Pipkin'},
+          // {text: 'Lou Todd', id: 'Lou Todd'},
+          // {text: 'Carol Beer', id: 'Carol Beer'},
+          // {text: 'Wicky Pollard', id: 'Wicky Pollard'},
```
mcbhenwood, line 28: Optional: usually preferable to remove commented code.

cypress/e2e_tests/specs/pokus.test.js
```diff
@@ -0,0 +1,43 @@
...
+        ],
+      }
+    })
+
+    // cy.get('.select2')
```
mcbhenwood, line 33: Optional: usually preferable to remove commented code.

cypress/e2e_tests/specs/pokus.test.js
```diff
@@ -0,0 +1,43 @@
...
+    cy.get('#select2-id_crm-container')
+      .click()
+    
+    // cy.get('.select2-results')
```
mcbhenwood, line 37: Optional: usually preferable to remove commented code.

requirements.txt
```diff
@@ -62,3 +62,4 @@ uritemplate==3.0.0
 urllib3==1.25.7
 wcwidth==0.1.7
 whitenoise==5.0.1
+django-autocomplete-light==3.5.0
```
mcbhenwood, line 65: Ideally this file would be ordered alphabetically

PR #123, opened 2020-07-09
Add enquiry stage to upload tool

rachaelcodes:
Adds the enquiry stage field to the csv upload tool.


rachaelcodes:
> When I upload a csv with the stage set as 'New' (not in all caps) or 'AWAITING RESPONSE' (not using an underscore) it doesn't accept it. It looks like this is because the only options are the choices in `ref_data.py`.
> 
> It looks like we'd either need to give a drop down of options in the csv template (not sure how possible this is) or change the ref data to accept more human-readable values?

Unfortunately, this seems to be the case for the other fields; it seems to be how this has been set up (which might not be a problem if the csv is outputted from another source). I'm going to ask Billy for a bit more info about the use of this form, because if it isn't, this is a bigger change.

PR #125, opened 2020-07-13
Truncate pagination

mforner13:
This PR adds logic to truncate the page numbers returned when a `GET` request is made to `enquiry-list`. This prevents all page numbers showing at the bottom of the enquiry list when there are lots of results. 

The logic itself mirrors what we have in other DIT systems (Data Hub and Market Access). When there are more than 6 pages worth of enquiries, blocks of 4 pages will be returned at a time (related to what page you're on), with the first and last pages always being displayed and other pages being truncated to just an ellipses `...`.

E.g. if there are 7 pages of results: page numbers would be `[1, 2, 3, 4, '...', 7]`.

I have added aria labels to the page numbers and tested this with a screenreader (VoiceOver).

It would be particularly useful to get feedback on the tests, and how the two response objects could possibly be combined.

**Before**
image

**After**
Screenshot 2020-07-13 at 12 02 10

### DIFF
--- a/app/enquiries/views.py
+++ b/app/enquiries/views.py
@@ -72,21 +72,27 @@ class PaginationWithPaginationMeta(PageNumberPagination):
     """
 
     def get_paginated_response(self, data):
-        return Response(
-            {
-                "next": self.get_next_link(),
-                "previous": self.get_previous_link(),
-                "count": self.page.paginator.count,
-                "current_page": self.page.number,
-                "results": data,
-                "filter_enquiry_stage": get_enquiry_field("enquiry_stage"),
-                "owners": models.Owner.objects.all().order_by("last_name"),
-                "query_params": self.request.GET,
+        response_data = {
+            "next": self.get_next_link(),
+            "previous": self.get_previous_link(),
+            "count": self.page.paginator.count,
+            "current_page": self.page.number,
+            "results": data,
+            "filter_enquiry_stage": get_enquiry_field("enquiry_stage"),
+            "owners": models.Owner.objects.all().order_by("last_name"),
+            "query_params": self.request.GET,
             "total_pages": len(self.page.paginator.page_range),
-                                               page_number))
-                          for page_number in self.page.paginator.page_range],
-            },
-            template_name="enquiry_list.html",
+            "pages": [{'page_number': page_number,
+                        'current': page_number == self.page.number,
+                        'link': replace_query_param(
+                                            self.request.get_full_path(),
+                                            'page',
+                                            page_number)}
+                        for page_number in self.page.paginator.page_range],
+
+            }
+        return Response(
+            truncate_response_data(response_data),
         )
 
     def post(self, request, format=None):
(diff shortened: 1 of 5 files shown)


app/enquiries/views.py
```diff
@@ -72,23 +72,26 @@ class PaginationWithPaginationMeta(PageNumberPagination):
     """
 
     def get_paginated_response(self, data):
+        pagination_data = {
```
peterhudec, line 75: I would name it `response_data` instead of `pagination_data` as only a portion of it is pagination related and it is used as the `data` argument to [`Response`](response).

BTW I'm wondering why we are using the `rest_framework.response.Response` instead of `django.shortcuts.render`?

app/enquiries/views.py
```diff
@@ -72,23 +72,26 @@ class PaginationWithPaginationMeta(PageNumberPagination):
...
+            "filter_enquiry_stage": get_enquiry_field("enquiry_stage"),
+            "owners": models.Owner.objects.all().order_by("last_name"),
+            "query_params": self.request.GET,
+            "total_pages": len(self.page.paginator.page_range),
+            "pages": [{'page_number': page_number,
+                        'current': page_number == self.page.number,
+                        'link': replace_query_param(self.request.get_full_path(),
+                                            'page',
+                                            page_number)}
+                        for page_number in self.page.paginator.page_range],
```
peterhudec, line 90: I would format it so that it fits within 80 characters per line.

app/enquiries/views.py
```diff
@@ -99,6 +102,56 @@ def post(self, request, format=None):
             return Response(serializer.data, status=status.HTTP_201_CREATED)
 
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    
+    def truncate_pagination_data(self, pagination_data, block_size=4):
```
peterhudec, line 106: This could be a _class method_ or just a function as you are not using `self`.

app/enquiries/views.py
```diff
@@ -99,6 +102,56 @@ def post(self, request, format=None):
...
+        """
+        Truncate the pagination links.
+
+        We don't want to show a link for every page if there are lots of page.
```
peterhudec, line 110: Typo in "...lots of page(s)"

app/enquiries/views.py
```diff
@@ -99,6 +102,56 @@ def post(self, request, format=None):
             return Response(serializer.data, status=status.HTTP_201_CREATED)
 
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    
+    def truncate_pagination_data(self, pagination_data, block_size=4):
+        """
+        Truncate the pagination links.
+
+        We don't want to show a link for every page if there are lots of page. 
+        This replaces page links which are less useful with an elipses ('...').
```
peterhudec, line 111: Typo in "elipses", should be "ellipsis".

app/enquiries/views.py
```diff
@@ -99,6 +102,56 @@ def post(self, request, format=None):
...
+
+        if len(pages) <= block_size:
+            return pagination_data
+
+        current_page_num = pagination_data["current_page"]
+        current_page_index = pagination_data["current_page"] - 1
+        first_page = pages[0]
+        last_page = pages[-1]
+
+        block_pivot = int(block_size / 2)
```
peterhudec, line 123: You can use the `//` operator which always returns an integer:
```python
block_pivot = block_size // 2
```

app/enquiries/views.py
mforner13: Ah so `PaginationWithPaginationMeta` is used as the `pagination_class` in `EnquiryListView`. It looks like it shouldn't actually be rendering a template itself, as that is handled in `EnquiryListView`. So I have removed the template as an argument to Response and everything still works fine.

PR #126, opened 2020-07-14
Convert owner filter field to <select> input

rachaelcodes:
Makes the owner filter a <select> input instead of a series of checkboxes - have checked with PM that is OK to now only allow one owner to be selected in a filter.

Test fixtures also updated to allow pagination to continue to be tested as before.


cypress/e2e_tests/specs/filter.test.js
```diff
@@ -77,6 +79,14 @@ const submitFilters = () =>
   cy.contains('Apply filters')
     .as('submit')
     .click()
+  
+const setOwner = id => 
+    cy.get('#owner__id')
```
peterhudec, line 84: It's better to not use IDs in cypress. We should mimick the way a human would use the page and they would select it by the label. Here it is nicely explained (we shoul actually start using this library):

- https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change
- https://testing-library.com/docs/dom-testing-library/intro

cypress/e2e_tests/specs/filter.test.js
```diff
@@ -77,6 +79,14 @@ const submitFilters = () =>
...
+    .select(id === UNASSIGNED ? 'Unassigned' : USERS[id])
+
+const assertOwnerSet = id =>
+  cy.get('#owner__id')
```
peterhudec, line 88: See the comment above

PR #130, opened 2020-07-14
Updates enquiry_text to uneditable text field

rachaelcodes:
Updates enquiry_text from CharField (with character limit) to TextField (no character limit).

In the edit form, provides the enquiry_text as reference, without ability to edit:
Screenshot 2020-07-14 at 16 09 11


app/enquiries/migrations/0013_auto_20200714_1226.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.0.7 on 2020-07-14 12:26
```
mforner13, line 1: The old team were leaving the auto-generated names for migration files, however I started naming the recent migrations based on what they're doing, as in Data Hub API. Better late than never :p

sass/_enquiry_edit.scss
```diff
@@ -1 +1,10 @@
 $govuk-assets-path: '/static/assets/';
+
+.enquiry-text {
+    @include govuk-responsive-margin(6, "bottom");
+
+    &__heading {
```
mforner13, line 6: Could use `class="govuk-label"` for the label instead?

rachaelcodes: I wasn't sure about this, as it's not a <label>. Though on the other hand, "govuk-label" would make it more futureproof if there are changes.

mforner13: A good point! Could it be a <label> instead of a div?

rachaelcodes: I thought maybe not because it's no longer surrounding an input (and I went for a div for the text instead of setting an input to 'disabled' or 'readonly' because it was more accessible: https://medium.com/@izzmo/anti-pattern-using-disabled-or-read-only-for-form-input-e2c24c669f5b )

app/enquiries/templates/enquiry_edit.html
```diff
@@ -135,7 +135,10 @@ <h3 class="govuk-heading-m">Enquiry overview</h3>
 
                 <h3 class="govuk-heading-m">Enquiry details</h3>
 
-                {% include 'snippets/enquiry_field_textarea.html' with instance=enquiry field="enquiry_text" %}
+                <div class="enquiry-text">
+                    <div class="govuk-label">Enquiry text</div>
+                    <div>{{ enquiry.enquiry_text }}</div>
```
peterhudec, line 140: Is the surrounding `<div>` needed? If so, shouldn't it be a `<p>`?

PR #133, opened 2020-07-15
Update cookie security flags

rachaelcodes:
A few updates to cookie security, using environment variables: 

- Updates the CSRF cookie to use the `HttpOnly` flag and the `Secure` flag where requested.
- Updates the session cookie to use the `Secure` flag. This setting would also be copied over to the messages cookie.
- Updates the sample environment variable file


app/settings/common.py
```diff
@@ -293,3 +293,8 @@
 ACTIVITY_STREAM_ENQUIRY_SEARCH_VALUE2 = env('ACTIVITY_STREAM_ENQUIRY_SEARCH_VALUE2')
 # key of the object that contains enquiry data
 ACTIVITY_STREAM_ENQUIRY_DATA_OBJ = env('ACTIVITY_STREAM_ENQUIRY_DATA_OBJ')
+
+# Settings for CSRF and Session cookies
+CSRF_COOKIE_SECURE = env('CSRF_COOKIE_SECURE', default=False)
+CSRF_COOKIE_HTTPONLY = env('CSRF_COOKIE_HTTPONLY', default=False)
+SESSION_COOKIE_SECURE = env('SESSION_COOKIE_SECURE', default=False)
```
peterhudec, line 300: Should't they default to `True` so that we accidentally don't deploy it insecure by forgetting to set them?

rachaelcodes: I'm copying the design in the API - I think it's default False in case people forget to set them locally, but I can double-check in Slack.

PR #137, opened 2020-07-21
Capitalise Owner name

mforner13:
When demoing the tool to users, when one user logged in on staging her name appeared as lower case across the pages.

To prevent this from happening, this PR adds the `title()` function to the Owner model's `__str__` method, meaning that whenever an owner object is referenced as a string, it will be capitalised. `title()` was used instead of `capitalise()` because this accounts for names with an apostrophe. E.g. O'Reilly.

In the `enquiry_item` template the Owner object itself isn't referenced, but rather the owner value from a dictionary of results, meaning that the `__str__` method can't be called. So the first and last name are 'titilised' directly in this template.

**To test**

Change a user's name to lower case in `test_users.json` and check that it is capitalised in the following areas:
- Filters
- The badge on an enquiry in the list
- Edit an enquiry
- View an enquiry
- Delete an enquiry

### DIFF
--- a/app/enquiries/templates/partials/enquiry_filters.html
+++ b/app/enquiries/templates/partials/enquiry_filters.html
@@ -14,7 +14,7 @@
         </option>
         {% for owner in owners %}
           <option value="{{ owner.id }}" {% query_params_value_selected owner.id 'owner__id' query_params ' selected'  %}>
-            {{ owner.get_full_name }}
+            {{ owner }}
           </option>
         {% endfor %}
       </select>
--- a/app/enquiries/models.py
+++ b/app/enquiries/models.py
@@ -31,11 +31,11 @@ class Enquirer(models.Model):
 
 class Owner(AbstractUser):
     """
-    Customer user model user by the app. Each Enquiry has an owner.
+    Customer user model used by the app. Each Enquiry has an owner.
     """
 
     def __str__(self):
-        return f"{self.first_name} {self.last_name}"
+        return f"{self.first_name.title()} {self.last_name.title()}"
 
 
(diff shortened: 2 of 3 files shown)


app/enquiries/templates/partials/enquiry_filters.html
```diff
@@ -14,7 +14,7 @@
         </option>
         {% for owner in owners %}
           <option value="{{ owner.id }}" {% query_params_value_selected owner.id 'owner__id' query_params ' selected'  %}>
-            {{ owner.get_full_name }}
+            {{ owner }}
```
mforner13, line 17: `__str__` does the same thing as `get_full_name` in this context, so I have used `__str__` rather than override the get_full_name method in the Owner class and capitalising that.

peterhudec: Is the `get_full_name` used elsewhere? I would get rid of it.

peterhudec: You could theoretically use `{{ owner|title }}` here instead of in the `__str__`. It would be more consistent with how you treat the first and last names. Also this seems to be about presentation which is the responsibility of the templates. But it also depends on whether you want the names to be capitalized in the CSV, because `__str__` is used there.

mforner13:
> Looks good, I didn't know about the `title` method. It however opens the question whether it also should be capitalized in the CSV.

Thanks for the comments @peterhudec 

`get_full_name` is a built in method so no need to remove it.

I have checked with Billy and he said if it's easy to capitalise names in CSV then do it, so will leave the `__str__` method as is :)

PR #138, opened 2020-07-22
Make error messages more human-readable

rachaelcodes:
Updates error messages to make them more human-readable

* updates the CSV upload error message strings:
Screenshot 2020-07-22 at 16 45 29

* creates a 'title_phrase' filter for the dictionary-based error messages in the `enquiry_detail` view
Screenshot 2020-07-22 at 16 45 43

* adds more human explanations to error messages stemming from data-hub upload task
* also brings the `sample_env` for local development up-to-date with api changes

### DIFF
--- a/app/enquiries/common/datahub_utils.py
+++ b/app/enquiries/common/datahub_utils.py
@@ -469,7 +469,7 @@ def dh_investment_create(request, enquiry, metadata=None):
     # check if the user is available in Data Hub
     user_details, error = dh_get_user_details(request, access_token)
     if error:
-        response["errors"].append({"referral_advisor": error})
+        response["errors"].append({"referral_advisor": "Error validating your identity in Data Hub"})
         return response
 
     dh_status = dh_enquiry_readiness(request, access_token, enquiry)
@@ -491,7 +491,7 @@ def dh_investment_create(request, enquiry, metadata=None):
         request, access_token, full_name, company_id
     )
     if error:
-        response["errors"].append({"contact_search": error})
+        response["errors"].append({"contact_search": f"Error while checking company contacts, {str(error)}"})
         return response
 
     primary = not existing_contacts
@@ -499,7 +499,7 @@ def dh_investment_create(request, enquiry, metadata=None):
         request, access_token, enquiry, company_id, primary=primary
     )
     if error:
-        response["errors"].append({"contact_create": error})
+        response["errors"].append({"contact_create": f"Error while creating a new company contact, {str(error)}"})
         return response
 
     contact_id = contact_response["id"]
--- a/app/enquiries/templates/enquiry_detail.html
+++ b/app/enquiries/templates/enquiry_detail.html
@@ -26,14 +26,14 @@
         {% if errors %}
         <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary">
             <h2 class="govuk-error-summary__title" id="error-summary-title">
-              There is a problem.
+                There is a problem.
             </h2>
             {% for error in errors %}
                 <div class="govuk-error-summary__body">
                 <ul class="govuk-list govuk-error-summary__list">
                     {% for key,value in error.items %}
                     <li>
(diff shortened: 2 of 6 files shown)


app/enquiries/common/datahub_utils.py
```diff
@@ -469,7 +469,7 @@ def dh_investment_create(request, enquiry, metadata=None):
     # check if the user is available in Data Hub
     user_details, error = dh_get_user_details(request, access_token)
     if error:
-        response["errors"].append({"referral_advisor": error})
+        response["errors"].append({"referral_advisor": f"Error validating your identity, {error}"})
```
mforner13, line 472: I'm not sure that showing the `error` here would be that useful? 

Screenshot 2020-07-23 at 11 45 02

If this error would come from the user not being in Data Hub, could we make that clearer perhaps? E.g. `Error validating your identity in Data Hub` The only thing is I'm not sure what they could do to resolve this apart from check they are definitely a Data Hub user.

app/enquiries/views.py
```diff
@@ -489,7 +493,8 @@ def post(self, request, *args, **kwargs):
         try:
             records = self.process_upload(file_obj)
         except Exception as err:
-            messages.add_message(request, messages.ERROR, str(err))
+            for msg in parse_error_messages(err):
```
adamledwards, line 496: I believe your trying to parse Django `ValidationError`
```python
   try:
            records = self.process_upload(file_obj)
   except ValidationError as err:
       for msg in parse_error_messages(err):
           messages.add_message(request, messages.ERROR, msg)
        return HttpResponseRedirect(self.ERROR_URL)  
    except Exception:
        messages.add_message(request, messages.ERROR, "Unexpected error")
        return HttpResponseRedirect(self.ERROR_URL)  
```
As a general observation the Exceptions are a bit too generic and if there is an expected error such as validation I think that ValidationError should be explicitly caught.
Slack me if you want more context

app/enquiries/templates/enquiry_detail.html
```diff
@@ -26,14 +26,14 @@
         {% if errors %}
         <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary">
             <h2 class="govuk-error-summary__title" id="error-summary-title">
-              There is a problem.
+                There is a problem.
```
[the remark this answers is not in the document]
rachaelcodes: This was to make the tabs line up.

app/enquiries/common/datahub_utils.py
rachaelcodes: I think this is also the error that comes up when the session is out of date, so maybe Peter's changes will make it clearer.

mforner13: As 'Referral Advisor' isn't a field the user can see, I think some reference to Data Hub would be useful. If you click on the link it doesn't take you anywhere as `#referral_adviser` isn't an id on the page.

rachaelcodes:
> nice does it work with the validationError?

@adamledwards  yes it does - thanks for the tip!

sample_env
```diff
@@ -11,12 +11,12 @@ DJANGO_SENTRY_DSN=sentry-dsn
 ENQUIRIES_PER_PAGE=10
 
 # Data Hub variables
-DATA_HUB_METADATA_URL=http://dh-api-mock:8001/metadata
-DATA_HUB_COMPANY_SEARCH_URL=http://dh-api-mock:8001/company/search
-DATA_HUB_CONTACT_SEARCH_URL=http://dh-api-mock:8001/contact/search
-DATA_HUB_CONTACT_CREATE_URL=http://dh-api-mock:8001/contact/create
-DATA_HUB_INVESTMENT_CREATE_URL=http://dh-api-mock:8001/investment/create
-DATA_HUB_ADVISER_SEARCH_URL=http://dh-api-mock:8001/adviser/search
+DATA_HUB_METADATA_URL=http://dh-api-mock:8001/v4/metadata
```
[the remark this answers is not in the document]
rachaelcodes: Have reverted this to fix the tests - looks like the '/' gets appended

PR #139, opened 2020-07-23
Set strict transport security header for prod

mforner13:
This PR adds the `strict-transport-security` header to all URLs in the application using Django's `SecurityMiddleware`. Having this header in place means that a user can only access the site at `https` not `http`. 

`SECURE_HSTS_SECONDS` sets the `max-age` value (the time for which a browser should remember that a site is only to be accessed using HTTPS), which I have set to 1 year. 
`SECURE_HSTS_INCLUDE_SUBDOMAINS` adds `includeSubdomains` to the header, which means that any subdomains are also only accessible via `https`.

This header is only needed in production, so I have set the default to true, and have set the env var `SET_HSTS_HEADERS` to false in dev, staging and the sample.env.

**To test**

I deployed this branch to dev before adding the new env var so you can test by:

- Going to the dev environment - https://enquiry-mgmt-dev.london.cloudapps.digital/enquiries/
- Checking that `strict-transport-security: max-age=31536000; includeSubDomains` is set in any request's Response headers


app/settings/common.py
```diff
@@ -218,7 +218,12 @@
 # enable the URL and turn on the Test Fixture API.
 ALLOW_TEST_FIXTURE_SETUP = env('ALLOW_TEST_FIXTURE_SETUP', default=None) == 'allow'
 
-# Sets the X-XSS-Protection HTTP header in older browsers
+# Set security related headers
+SET_HSTS_HEADERS = env('SET_HSTS_HEADERS', default=True)
```
rachaelcodes, line 222: So the pattern in DataHub API (rightly or wrongly) seems to be to set the default for these kinds of security settings to False - I think to mimic the Django default or to maybe to make it easier in dev (or to only have to set the env variable in prod): e.g.
`CSRF_COOKIE_SECURE = env('CSRF_COOKIE_SECURE', default=False)`

When I asked, I couldn't get a particular reason why we're doing this, but if you want to switch it round, it might also be worth changing the three COOKIE settings to also have a `default=True`

mforner13: I asked Mark about this one and he said that it would be better to default it to `True` so that it doesn't require any setting in production, and then set it as `False` everywhere else. I'll check with him whether it should be the same for cookies.

PR #140, opened 2020-07-23
Add CSV renderer to EnquiryListView

peterhudec:
Fixes a bug when the _Download all_ button didn't respect the currently set filters.


peterhudec:
> Does `rest_framework_csv` need to be added to `requirements.txt`?

Yes, it's added. The name is a bit different though: `djangorestframework-csv==2.1.0`.
https://github.com/uktrade/enquiry-mgmt-tool/pull/140/files#diff-b4ef698db8ca845e5845c4618278f29aR21

peterhudec:
> As the data grown Django will run out of memory or the server may timeout waiting for a response.
> The I can see its using `models.Enquiry.objects.all()` you will have to stream the response using a generator function.
> https://docs.djangoproject.com/en/3.0/howto/outputting-csv/#streaming-csv-files
> https://github.com/mjumbewu/django-rest-framework-csv/blob/master/rest_framework_csv/renderers.py#L196
> 
> My ideal solution would be to create your own renderer and use Django streaming one but its up to you.

Good point, there is also the issue with this is caching i.e. we don't want to recreate a huge CSV every time someone wants to load it. But don't want to address this just now as the current priority is to get it to production ASAP.

peterhudec:
> Looks good! Is it possible to add a test to check the expected enquiries are in the csv after filtering?
I'll have a look at it.

PR #142, opened 2020-07-27
Setup middleware tests

mforner13:
This PR sets up tests for middleware in `test_middleware.py` and checks tests that: 

1. The correct middleware are in settings
2. That the security setting `SECURE_BROWSER_XSS_FILTER` is turned on 
3. That the custom middleware `add_cache_control_header_middleware` adds the correct response header values

Some of these things were previously tested in `test_views.py` and have been moved to one file for clarity.

The tests don't check that some settings e.g. the `SECURE_HSTS` and `COOKIE` settings are turned on because they rely on an environment variable being `True`, which will vary between environments.

### DIFF
--- a/app/enquiries/tests/test_middleware.py
+++ b/app/enquiries/tests/test_middleware.py
@@ -1,0 +1,33 @@
+from django.urls import reverse
+
+import app.enquiries.tests.utils as test_utils
+from app.settings import common
+
+def test_correct_middleware_exists():
+  assert common.MIDDLEWARE == [
+    'django.middleware.security.SecurityMiddleware',
+    'app.middleware.add_cache_control_header_middleware',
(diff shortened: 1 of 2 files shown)


app/enquiries/tests/test_middleware.py
```diff
@@ -0,0 +1,34 @@
+from django.urls import reverse
+
+import app.enquiries.tests.utils as test_utils
+from app.settings import common
+
+headers = {"HTTP_CONTENT_TYPE": "text/html", "HTTP_ACCEPT": "text/html"}
```
elcct, line 6: This is only used by one test, so maybe it's good to move it to its method.

mforner13: Good point! Have moved it

PR #145, opened 2020-07-29
Mark non-required fields as optional

rachaelcodes:
Adds an 'optional' flag to fields that can be left blank when submitting to Data Hub.

Screenshot 2020-07-24 at 12 26 55

* Creates a new label snippet and filter for identifying optional fields (note that some fields may be blank in the model, but need to be submitted to DH. These are NOT marked as optional)
* Creates new filter for identifying select fields that can be left as the (----) default
* Also adds additional error handling for the DH request


app/enquiries/views.py
```diff
@@ -300,7 +300,7 @@ def post(self, request, *args, **kwargs):
         enquiry = context["enquiry"]
 
         create_response = dh_investment_create(request, enquiry)
-        if create_response.get("result") and create_response["result"]["id"]:
+        if create_response.get("result") and create_response["result"].get("id"):
```
mforner13, line 303: Why did you chose `.get("id")` over `["result"]["id"]`?

app/enquiries/templates/snippets/summary_list_key_value.html
```diff
@@ -1,7 +1,7 @@
 {% load enquiries_extras %}
 <div class="govuk-summary-list__row">
     <dt class="govuk-summary-list__key">
-        {{ instance|get_field_verbose_name:field }}
+        {{ instance|get_field_verbose_name:field }} {% if instance|is_optional:field %} (optional){% endif %}
     </dt>
     <dd class="govuk-summary-list__value"  id="{{ field }}">{% if instance|get_field_value:field == None %}{{ "----" }}{% else %}{{ instance|get_field_value:field }}{% endif %}</dd>
 </div>
```
mforner13, line 7: Missing new line at end

app/enquiries/templates/snippets/enquiry_field_textarea.html
```diff
@@ -1,8 +1,6 @@
 {% load enquiries_extras %}
 <div class="govuk-form-group">
-    <label class="govuk-label" for="{{ field }}">
-        {{ instance|get_field_verbose_name:field }}
-    </label>
+    {% include 'snippets/enquiry_field_label.html' with instance=instance field=field %}
     <textarea class="govuk-textarea" id={{ field }} name={{ field }}
         rows="5">{{ instance|get_field_value:field }}</textarea>
 </div>
```
mforner13, line 6: Missing new line at end

app/enquiries/templates/snippets/enquiry_field_select.html
```diff
@@ -2,6 +2,7 @@
 <div class="govuk-form-group">
     <label class="govuk-label" for="{{ field }}">
         {{ instance|get_field_verbose_name:field }}
+        {% if field|can_be_default %} (optional){% endif %}
```
mforner13, line 5: Could you use the `enquiry_field_label` snippet here?

app/enquiries/views.py
peterhudec, line 303: This is hairy condition could be simplified to:
```python
if create_response.get("result", {}).get("id"):
    # ...
```

app/enquiries/common/datahub_utils.py
```diff
@@ -516,6 +516,9 @@ def dh_investment_create(request, enquiry, metadata=None):
 
     try:
         result = dh_request(request, access_token, "POST", url, payload)
+        
+        result.raise_for_status()
```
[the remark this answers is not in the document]
rachaelcodes: Is it worth doing two excepts? One for HTTP error, one for other ones?

app/enquiries/templates/snippets/enquiry_field_select.html
rachaelcodes: The logic for this filter is a bit different from the one used in the snippet - unlike the other `<input>`s, the `<select>`s can't be blank. Rather than having a convoluted (X and Y) or Z logic, I created two separate filters.

app/enquiries/views.py
rachaelcodes: If create_response["result"] exists but didn't have an ["id"] key, it would error - this was a safer way of checking if it existed. Peter's also suggested a cleaner way to write this logic.

PR #148, opened 2020-07-31
Replace primary sector with ist sector in template

mforner13:
## Description of change

We have received feedback from the users that when they are uploading enquiries via a CSV, these enquiries would have an `IST Sector` not a `Primary sector`, as the primary sector is only relevant for submitting an enquiry to Data Hub.

This PR replaces `Primary sector` with `IST Sector` in the CSV template as well as in the factory to create a csv row and the management command for exporting a CSV of all enquiries. 

I have also corrected the reference data class `IstSector` to be `ISTSector` as IST are initials.

## Test instructions

When you download the import template, there should now be an `ist_sector` field instead of `primary_sector`

### DIFF
--- a/app/enquiries/utils.py
+++ b/app/enquiries/utils.py
@@ -107,7 +107,7 @@ def generate_import_template(file_obj):
         ref_data.EnquiryStage,
         ref_data.HowDidTheyHear,
         ref_data.InvestmentReadiness,
-        ref_data.IstSector,
+        ref_data.ISTSector,
         ref_data.RequestForCall,
         ref_data.MarketingChannel,
     ]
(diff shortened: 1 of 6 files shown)


app/enquiries/utils.py
```diff
@@ -115,7 +115,7 @@ def generate_import_template(file_obj):
         ref_data.EnquiryStage,
         ref_data.HowDidTheyHear,
         ref_data.InvestmentReadiness,
-        ref_data.PrimarySector,
+        ref_data.IstSector,
```
peterhudec, line 118: Shouldn't it be `ISTSector`?

> When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError - [PEP-8](https://www.python.org/dev/peps/pep-0008/#id5)

mforner13: A good point! That's how it is in the ref data currently, but I will update it :)

mforner13: @peterhudec this is now changed

PR #149, opened 2020-07-31
Add the SESSION_COOKIE_AGE setting

peterhudec:
## Description of change

This PR sets the [`SESSION_COOKIE_AGE`](https://docs.djangoproject.com/en/3.0/ref/settings/#session-cookie-age) Django setting to 9 hours and allows it to be overriden with an env var of the same name. This is a measure against the SSO access token expiration, i.e. the session expires sooner than the token (10 hours), which redirects the user to `/auth/login/` and refreshes both the access token and session.

## Test instructions

1. Set the `SESSION_COOKIE_AGE` in your `.env` file to a short time e.g. a minute (`SESSION_COOKIE_AGE=60`).
2. Visit the app from your browser
3. Clear all cookies
4. Refresh the page
5. Verify that there was a cookie `sessionid` created with _Expires/Max Age_ set to a time and date corresponding to the aforementioned setting
6. Wait till the expiration period is over
7. Click on one of the enquiries
8. You should be redirected to `/auth/login`, then to the SSO service and then back to the home page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,18 @@ Or in app/settings/*
 In which case, it will redirect to Django admin page for login so a superuser
 needs to be created first.
 
+#### OAuth Access Token Refreshment
+
+_Access tokens_ issued by Staff SSO have expiration time of 10 hours to
+just about outlive a user's working time. In order to always have a valid
+_access token_ this app limits the user's session to 9 hours. When the session
+expires, the user will be automatically redirected to `/auth/login` which will
+refresh both the session and the _access token_ and allows the user to use the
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -70,6 +70,18 @@ Or in app/settings/*
 In which case, it will redirect to Django admin page for login so a superuser
 needs to be created first.
 
+#### OAuth Access Token Refreshment
+
+_Access tokens_ issued by Staff SSO have expiration time of 10 hours to
```
mforner13, line 75: *an expiration time

PR #152, opened 2020-08-03
Add date_added field to enquiry model

rachaelcodes:
## Description of change
Gives the option of users adding a custom 'date_received' date which can be different from the date the enquiry is created.

As this is a custom option, the value shown to the user is a property on the model called 'received', which returns the 'date_received' value if that exists, otherwise defaults to the date created.

Also updates filters to use this custom search and updates tests.

## Test instructions
Run the app: http://localhost:8000/enquiries/46/ should have a 'Date Received' date in 2017 (not 2019, which is when 'created' is set in the test data).

Alternatively, visit: http://localhost:8000/enquiries/import, and attempt uploading an enquiry with a 'date_received' date set to a date other than the one when you upload it (use the format: YYYY-MM-DD); the 'Date Received' should display that date.


app/enquiries/fixtures/test_enquiries.json
```diff
@@ -1967,5 +1967,48 @@
...
+            "client_relationship_manager": "Data Hub user 1",
+            "project_code": "42901",
+            "date_added_to_datahub": "2020-02-03",
+            "datahub_project_status": "ACTIVE",
+            "project_success_date": "2022-02-03",
+            "enquirer": 1,
+            "date_received": "2017-01-01T00:13:31"
+        }
     }
 ]
```
mforner13, line 2014: Missing new line

app/enquiries/serializers.py
```diff
@@ -81,6 +82,7 @@ class EnquiryDetailSerializer(serializers.ModelSerializer):
     date_added_to_datahub = serializers.DateField(format="%d %B %Y")
     datahub_project_status = serializers.CharField(source="get_datahub_project_status_display")
     project_success_date = serializers.DateField(format="%d %B %Y")
+    received = serializers.DateTimeField(format="%d %B %Y")
```
mforner13, line 85: When fetching an enquiry would both the received and date_received fields be returned? If so, I think just having received would avoid duplication.

app/enquiries/models.py
```diff
@@ -43,11 +43,18 @@ def __str__(self):
 class Enquiry(TimeStampedModel):
     """
     Model for investment Enquiry
+
+    Includes an optional 'date_received' field to allow a differentiation between when the enquiry
+    is uploaded to the system and when it was received by our teams.
```
mforner13, line 48: You can justify the lines in the docstring when it's just being used as a comment as the string won't be rendered anywhere

app/enquiries/views.py
```diff
@@ -221,13 +221,33 @@ def filter_owner_id(self, queryset, name, value):
...
+            )
+        return queryset.filter(q)
 
-    def filter_created_gt(self, queryset, name, value):
-        created = datetime.combine(value, datetime.min.time())
-        return queryset.filter(created__gt=created)
+    def filter_received_gt(self, queryset, name, value):
+        """
+        Reproduces the 'received' Enquiry Model property for a 'greater-than' search.
+
```
mforner13, line 241: Same comment as above for the docstring

app/enquiries/views.py
```diff
@@ -221,13 +221,33 @@ def filter_owner_id(self, queryset, name, value):
 
         return queryset.filter(q)
 
-    def filter_created_lt(self, queryset, name, value):
-        created = datetime.combine(value, datetime.min.time())
-        return queryset.filter(created__lt=created)
+    def filter_received_lt(self, queryset, name, value):
+        """
+        Reproduces the 'received' Enquiry Model property for a 'less-than' search.
```
peterhudec, line 226: I would change the wording "Reproduces ..." to something like "Returns a queryset only with entities having the \`\`received\`\` date less than \`\`value\`\`"

We should also use proper [Sphinx markup](https://www.sphinx-doc.org/en/master/) in docstrings e.g. the double backticks.

Here's a [Sphinx cheat sheet](http://openalea.gforge.inria.fr/doc/openalea/doc/_build/html/source/sphinx/rest_syntax.html)

rachaelcodes: Ah nice - thanks for the cheat sheet

PR #156, opened 2020-08-04
Handle user submitting 'Import Enquiries' form without selecting a file.

rachaelcodes:
## Description of change
Currently, if the user attempts to submit the 'Import Enquiries' form without selecting a file, there is a 500 error. This adds a clear error message instead.

## Test instructions
Visit http://localhost:8000/enquiries/import .
Click 'Upload File' without selecting a file to add.
You should see the following:
Screenshot 2020-08-04 at 16 00 38

### DIFF
--- a/app/enquiries/tests/test_views.py
+++ b/app/enquiries/tests/test_views.py
@@ -439,9 +439,32 @@ class EnquiryViewTestCase(test_utils.BaseEnquiryTestCase):
             # clear entries for next mime type run
             Enquiry.objects.all().delete()
 
+    def test_enquiry_import_post_blank(self):
+        """
+        Test an unsuccessful import, where no file is provided
+        """
+        initial_count = Enquiry.objects.count()
+
+        response = self.client.post(
+            reverse("import-enquiries"), {"enquiries": ""}, follow=True
+        )
+
+        soup = BeautifulSoup(response.content, "html.parser")
+        error_message = soup.select(".error")[0].text
+
+        self.assertEqual(error_message, "No file was selected. Choose a file to upload.")
+
+        # test atomic transactions
+        final_count = Enquiry.objects.count()
+        self.assertEqual(
+            final_count,
+            initial_count,
+            f"atomic transaction should prevent any records being created. Found: {final_count}",
+        )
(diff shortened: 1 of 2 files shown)


app/enquiries/views.py
```diff
@@ -494,9 +494,15 @@ def post(self, request, *args, **kwargs):
...
+            messages.error(
+                self.request,
+                "No file was selected. Please choose a file to upload.",
+            )
+            return HttpResponseRedirect(reverse("import-enquiries"))
+
         if not (
-            file_obj
-            and file_obj.name.endswith(".csv")
+            file_obj.name.endswith(".csv")
```
peterhudec, line 505: Not sure about this, a file extension only has a meaning to distinguish different file formats, but we only accept CSV here so it shouldn't matter what extension it has if at all. Also this won't help if they upload a missformatted `*.csv` file.

app/enquiries/views.py
```diff
@@ -494,9 +494,15 @@ def post(self, request, *args, **kwargs):
         enquiries_key = "enquiries"
 
         file_obj = request.FILES.get(enquiries_key)
+        if not file_obj:
+            messages.error(
+                self.request,
+                "No file was selected. Please choose a file to upload.",
```
mforner13, line 500: In the design system they recommend not using please in error messages as it 'implies a choice' https://design-system.service.gov.uk/components/error-message/. I'm sure it wouldn't make too much of a difference on this particular page, but could change to just `No file was selected. Choose a file to upload.` to be on the safe side.

rachaelcodes: Thanks for the link!

app/enquiries/tests/test_views.py
```diff
@@ -439,6 +439,29 @@ def test_enquiry_import_post_success(self):
             # clear entries for next mime type run
             Enquiry.objects.all().delete()
 
+    def test_enquiry_import_post_blank(self):
+        """
+        Test a unsuccessful import, where no file is provided
```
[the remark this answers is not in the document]
rachaelcodes: urghhhhh

PR #157, opened 2020-08-05
Disable 'Date Added to DataHub' on enquiry edit form

rachaelcodes:
## Description of change
Updates the Enquiry edit form so that the 'Date Added to Data Hub' field is information-only. 

## Test instructions
Select an enquiry, select 'edit', and see:
Screenshot 2020-08-05 at 11 19 16

### DIFF
--- a/app/enquiries/templatetags/enquiries_extras.py
+++ b/app/enquiries/templatetags/enquiries_extras.py
@@ -3,12 +3,11 @@ import os
 from django.conf import settings
 from django import template
 
-from app.enquiries import models
-from app.enquiries import serializers
+from app.enquiries import models, ref_data, serializers
 
 register = template.Library()
 
-dh_required_fields = [
+dh_strict_fields = [
     "client_relationship_manager",
     "project_name",
     "project_description",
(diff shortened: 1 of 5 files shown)


app/enquiries/templatetags/enquiries_extras.py
```diff
@@ -71,6 +71,21 @@ def get_dh_company_url(enquiry):
...
+def get_dh_date_added(enquiry):
+    if enquiry.date_added_to_datahub:
+        return enquiry.date_added_to_datahub.strftime("%d %B %Y")
+
+    if (
+        enquiry.datahub_project_status != ref_data.DatahubProjectStatus.DEFAULT
+    ) or (
+        enquiry.enquiry_stage == ref_data.EnquiryStage.ADDED_TO_DATAHUB
+    ):
+        return "Date not recorded"
```
peterhudec, line 84: I'd find it more readable and _pythonic_ if the two conditions were named. Also, It the logic could be expressed with just the `and` operator:
```python
default_status = enquiry.datahub_project_status == DatahubProjectStatus.DEFAULT
added_to_dh_stage = enquiry.enquiry_stage == EnquiryStage.ADDED_TO_DATAHUB

if default_status and added_to_dh_stage:
    return "Date not recorded"
```

PR #166, opened 2020-08-12
Truncate enquiry text in list view

mforner13:
## Description of change

Currently when viewing a list of enquiries, the whole enquiry text is showing. This can often be extremely long and so takes up a lot of the page.

This PR truncates the enquiry text when it is longer than 250 characters. 

I have used a custom template filter for this as Django's built-in `truncatechars` will cut off a string mid-word.

I have adjusted two existing enquiry fixtures in order to test this functionality.

## Test instructions

- Search for this company - ABC Electronics Co. (long text)
- While still in the list view, check that the enquiry text is cut off after 250 characters but that the final word intact


app/enquiries/templatetags/enquiries_extras.py
```diff
@@ -166,3 +166,18 @@ def query_params_value_selected(value, param_key, query_params, text="selected")
...
+    """
+    Truncates a string after a specified character length
+    but does not cut off mid-word.
+    """
+    if len(value) > max_length:
+        truncd_val = value[:max_length]
+        if value[max_length] != " ":
+            truncd_val = truncd_val[:truncd_val.rfind(" ")]
+        return truncd_val + "..."
+    else:
```
ambrozic, line 182: this `else` is not necessary

ambrozic:
would adding backend unit tests for `templatetags` (on top of existing frontend ones) make sense?

app/enquiries/templatetags/enquiries_extras.py
mforner13: Fixed :)

mforner13:
> would adding backend unit tests for `templatetags` (on top of existing frontend ones) make sense?

Have discussed and agreed that a unit test is not necessary as the e2e test covers all cases.

PR #167, opened 2020-08-13
Add admin models

peterhudec:
## Description of change

Enables editing the `Enquiry`, `Owner` and `Enquirer` models in Django admin.

## Test instructions

1. Go to `/admin/`
2. You should see the three sections there

Screen Shot 2020-08-05 at 5 04 39 PM
Screen Shot 2020-08-05 at 5 04 57 PM
Screen Shot 2020-08-05 at 5 05 09 PM
Screen Shot 2020-08-05 at 5 05 25 PM

### DIFF
--- a/app/enquiries/admin.py
+++ b/app/enquiries/admin.py
@@ -1,1 +1,33 @@
-# Register your models here.
+from django.contrib import admin
+
+from app.enquiries.models import Enquiry, Enquirer, Owner
+
+
+def all_fields_of(model, exclude=()):
+    return [field.name for field in
+            model._meta.fields if field.name not in exclude]
+
+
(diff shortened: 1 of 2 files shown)


app/enquiries/admin.py
```diff
@@ -1 +1,33 @@
-# Register your models here.
+from django.contrib import admin
+
+from app.enquiries.models import Enquiry, Enquirer, Owner
+
+
+def all_fields_of(model, exclude=()):
```
ambrozic, line 6: even if `exclude` default value is immutable tuple it would be more correct if default value was `None` and then you would test it like: 

```for f in fields if f.name not in (exclude or [])```

PR #169, opened 2020-08-14
Fix missing Content-Disposition header for CSV download

peterhudec:
## Description of change

Fixes an issue when the _Download all_ button rendered the CSV content in the browser instead of offering a file for download.

## Test instructions

1. Go to `/` or `/enquiries/`
2. Apply some filters (optional)
3. Click on the _Download all_ button
4. A file download dialogue with the suggested file name `rtt_enquiries_export.csv` should be opened, or a file with the same name should be downloaded, depending on your browser setup

### DIFF
--- a/app/enquiries/views.py
+++ b/app/enquiries/views.py
@@ -26,7 +26,7 @@ from rest_framework.renderers import TemplateHTMLRenderer
 from rest_framework.response import Response
 from rest_framework.utils.urls import replace_query_param
 from rest_framework.views import APIView
-from rest_framework_csv.renderers import CSVRenderer
+from rest_framework_csv.renderers import PaginatedCSVRenderer
 
 from app.enquiries.common.datahub_utils import dh_investment_create
 from app.enquiries import forms, models, serializers, utils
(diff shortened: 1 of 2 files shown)


app/enquiries/views.py
```diff
@@ -264,19 +264,21 @@ class EnquiryListView(LoginRequiredMixin, ListAPIView):
...
-    def paginator(self):
-        """This method override is here to disable pagination for the csv format"""
+    def finalize_response(self, *args, **kwargs):
+        """Handles the ``Content-Disposition`` header of a ``format=csv`` request"""
+        response = super().finalize_response(*args, **kwargs)
+
         if self.request.query_params.get("format") == "csv":
-            self._paginator = None
-        return super().paginator
+            response['Content-Disposition'] = "attachment; filename=rtt_enquiries_export.csv"
```
ambrozic, line 279: could we use `EXPORT_OUTPUT_FILE_SLUG` and `EXPORT_OUTPUT_FILE_EXT` here?

PR #170, opened 2020-08-14
Autopopulate date_received field when not provided

mforner13:
## Description of change

When an enquiry is uploaded via CSV, rather than ingested through Activity stream, the date it was received is often not the same as the date the enquiry is created in the tool. In order to return the date received, there is currently a `received` property on the Enquiry model which will return `date_received` if populated and `created` if not. As we are now implementing sorting and as the app scales, making queries on this property is proving complex.

As a result, this PR automatically populates the `date_received` field when an enquiry is created.

If an enquiry is uploaded via CSV and there is no value in the `date_received` column, that value will be set to `datetime.now()`.

If an enquiry is ingested via Activity Stream, the `date_received` value will be set to the `published` value from Activity stream (which is the moment the original form is filled out on the invest in great website.

TODO:

1. Next: write a management command to populate all existing enquiries which don't have a value for `date_received` with their date created
2. Run this management command 
3. Once all enquiries in production have a value for `date_received`, remove the `received` property from the enquiry model and replace uses of it with references to the `date_received` field (e.g. in filters)
4. Implement the sortby feature which will at this point be able to order enquiries by `date_received`

## Test instructions

The app is only integrated with activity stream in production, but you can test that the date_received field is populated when an enquiry is uploaded via CSV by:
- Uploading an enquiry without a date_received ([google doc here](https://docs.google.com/spreadsheets/d/1Bng4CbGYJPewKBy_NhY00fEt3zcTXO90/edit#gid=1423515304))
- Open a shell in the docker container (`docker exec -it enquiry-mgmt-tool_app_1 bash`
- Open a django shell, find the enquiry you just created and check that it has a value for `date_received`

### DIFF
--- a/app/enquiries/tests/test_utils.py
+++ b/app/enquiries/tests/test_utils.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 from django.test import Client, TestCase
 from faker import Faker
+from freezegun import freeze_time
 
 import app.enquiries.tests.utils as test_utils
 from app.enquiries import models, utils
@@ -45,6 +46,7 @@ class EnquiryViewTestCase(TestCase):
         exists = models.Enquiry.objects.filter(**qs_args).exists()
         self.assertTrue(exists)
 
+    @freeze_time()
     def test_util_row_to_enquiry_no_date_received(self):
         """
         Tests that when an enquiry is uploaded without date_received
--- a/app/enquiries/tests/test_utils.py
+++ b/app/enquiries/tests/test_utils.py
@@ -1,10 +1,14 @@
+from datetime import datetime
 from django.test import Client, TestCase
 from faker import Faker
 
 import app.enquiries.tests.utils as test_utils
 from app.enquiries import models, utils
 
-from app.enquiries.tests.factories import create_fake_enquiry_csv_row
+from app.enquiries.tests.factories import (
+    create_fake_enquiry_csv_row,
+    create_fake_enquiry_csv_row_no_date_received
+)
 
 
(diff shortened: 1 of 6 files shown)


app/enquiries/tests/test_utils.py
```diff
@@ -41,6 +45,15 @@ def test_util_row_to_enquiry(self):
...
         self.assertTrue(exists)
 
+    def test_util_row_to_enquiry_no_date_received(self):
+        """
+        Tests that when an enquiry is uploaded without date_received
+        populated, the field is populated with the current date.
+        """
+        csv_data = create_fake_enquiry_csv_row_no_date_received()
+        enquiry = utils.row_to_enquiry(csv_data)
+        assert enquiry.date_received.date() == datetime.now().date()
```
ambrozic, line 55: what if `enquiry` object is created before midnight and assertion runs after midnight? :stuck_out_tongue:

mforner13: A very good point! I could set today's date as a variable before or after assigning `enquiry`, but it could also hit midnight between those two points 😱  What do you think?

ambrozic: i think it is handy just to have [freezegun](https://pypi.org/project/pytest-freezegun/) at hand and simply wrap tests like this in a context manager

mforner13: @peterhudec @rachaelcodes really sorry have got to sign off now, would you be able to pick up adding freezegun in this PR next week? :)

PR #171, opened 2020-08-17
Add management command to fill date_received fields

rachaelcodes:
## Description of change
We are updating the `date_received` value of an enquiry to default to the `created` value where it is not explicitly provided. This management command is a task to add this missing information for existing data.

## Test instructions

- Run this command locally with the default test data (e.g. `docker exec -it enquiry-mgmt-tool_app_1 python manage.py fill_date_received` )
- Open up a django shell and confirm that the data has been updated

### DIFF
--- a/app/enquiries/management/commands/fill_date_received.py
+++ b/app/enquiries/management/commands/fill_date_received.py
@@ -1,0 +1,24 @@
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from app.enquiries.models import Enquiry
+
+
+class Command(BaseCommand):
+    """
+    Command is for one-off use to populate the date_received value for existing enquiries,
+    replacing empty values with the date the record was created
+    """
+
+    help = "this command replaces empty date_received values for enquiries with the date created"
+
+    def handle(self, *args, **options):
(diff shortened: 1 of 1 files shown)


app/enquiries/management/commands/fill_date_received.py
```diff
@@ -0,0 +1,24 @@
...
+        print('Replacing empty date_received values')
+
+        entries = Enquiry.objects.select_for_update().filter(date_received=None)
+        with transaction.atomic():
+            for entry in entries:
+                entry.date_received = entry.created
+                entry.save()
+
+        [print(entry.date_received) for entry in Enquiry.objects.all()]
+        print(f'Updated {len(entries)} enquiries')
```
peterhudec, line 24: I would do the print in the for loop above. List comprehensions are for transforming data, ideally without side effects. Also, this would keep a listo of `None`s in with as many items as there are enquiries untill it's garbage collected.

rachaelcodes: The idea was to demonstrate that the values had actually changed - but on second thoughts, I have an idea of how to do that within the same loop

PR #173, opened 2020-08-19
Add sort by functionality

rachaelcodes:
## Description of change
Adds an option to sort enquiries by different properties
As a default, enquiries will be sorted by date received, with the newest first.

sort select

## Test instructions
- When running locally, you will need to update SASS by running: `sh ./bootstrap.sh` to see expected styling
- Click on the 'Sort by' <select> to choose different options for sorting enquiry results; this sorting should be continued through to different pages of results

### DIFF
--- a/app/enquiries/templates/enquiry_list.html
+++ b/app/enquiries/templates/enquiry_list.html
@@ -30,10 +30,31 @@
                                 </div>
                             </div>
                         </div>
-                        <div class="collection-header__row">
+                        <div class="collection-header__row collection-header__row-no-border">
                             <div class="header-page-number-container">
                                 <span class="header-page-number">Page {{ current_page }} of {{ total_pages }}</span>
                             </div>
+                            <div class="sortby-container">
+                                <label for="sortby" class="govuk-label sortby-label">
+                                    Sort by
+                                </label>
+                                <select name="sortby" class="govuk-select sortby-select" onchange="window.location.href='{% url 'enquiry-list' %}?sortby=' + this.value">
+                                    <option value="name" {% if request.query_params.sortby == "name" %} selected {% endif %}>
+                                        Company name: A-Z
+                                    </option>
+                                    <option value="updated" {% if request.query_params.sortby == "updated" %} selected {% endif %}>
+                                            Most recently updated
+                                    </option>
+                                    <option value="received:asc" {% if request.query_params.sortby == "received:asc" %} selected {% endif %}>
+                                            Date received: old-new
+                                    </option>
+                                    <option value="received:desc" {% if request.query_params.sortby == "received:desc" %} selected {% endif %}>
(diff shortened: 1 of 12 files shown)


ambrozic (approved):

app/enquiries/templates/enquiry_list.html
```diff
@@ -30,10 +30,31 @@ <h2 class="govuk-heading-l">Enquiries</h2>
...
-                        <div class="collection-header__row">
+                        <div class="collection-header__row collection-header__row-no-border">
                             <div class="header-page-number-container">
                                 <span class="header-page-number">Page {{ current_page }} of {{ total_pages }}</span>
                             </div>
+                            <div class="sortby-container">
+                                <label for="sortby" class="govuk-label sortby-label">
+                                    Sort by
+                                </label>
+                                <select name="sortby" class="govuk-select sortby-select" onchange="window.location.href='{% url 'enquiry-list' %}?sortby=' + this.value">
```
rafenden, line 41: Perhaps, there is a better way of handling these events?

app/enquiries/templates/enquiry_list.html
```diff
@@ -30,10 +30,31 @@ <h2 class="govuk-heading-l">Enquiries</h2>
...
+                                    <option value="company_name" {% if request.query_params.sortby == "company_name" %} selected {% endif %}>
+                                        Company name: A-Z
+                                    </option>
+                                    <option value="-modified" {% if request.query_params.sortby == "-modified" %} selected {% endif %}>
+                                            Most recently updated
+                                    </option>
+                                    <option value="date_received" {% if request.query_params.sortby == "date_received" %} selected {% endif %}>
+                                            Date received: old-new
```
rafenden, line 49: Shall we use `oldest first` instead? 

`old-new` might not be clear for some users.

app/enquiries/templates/enquiry_list.html
```diff
@@ -30,10 +30,31 @@ <h2 class="govuk-heading-l">Enquiries</h2>
...
+                                    </option>
+                                    <option value="-date_received" {% if request.query_params.sortby|is_default_sort %} selected {% endif %}>
+                                        Date received: new-old
```
rafenden, line 52: Shall we use `newest first` instead?

`new-old` might not be clear for some users.

app/enquiries/templates/enquiry_list.html
rachaelcodes: I've changed it to use editing the 'location.search' setting, as Peter did with the CSV download button, which also means that filters are kept with the sorting. 

I'm not a massive fan of the inline scripts, but we're not really using JS elsewhere in the project.

peterhudec (approved):
Looks good, only the `sass/_enquiry_list.scss` has forgotten git conflict markers.

PR #175, opened 2020-08-20
Update CSV field choices

rachaelcodes:
## Description of change
Initially, too many fields were being returned to the users when they downloaded the Enquiry list summary CSV. Now only the expected fields are returned.

## Test instructions
From the homepage, click 'Download all'.
The first field displayed on the downloaded CSV should be: client_relationship_manager
The following fields should not be included:

- anonymised_project_description
- company_hq_address
- dh_assigned_company_name
- dh_company_address
- dh_company_id
- dh_company_number
- dh_duns_number
- enquirer.id
- id
- investor_involvement_level
- modified
- new_existing_investor
- organisation_type
- owner.date_joined
- owner.email
- owner.id
- owner.is_active
- owner.is_staff
- owner.is_superuser
- owner.last_login
- owner.password
- owner.username
- project_description
- specific_investment_programme

### DIFF
--- a/app/enquiries/tests/test_views.py
+++ b/app/enquiries/tests/test_views.py
@@ -562,6 +562,15 @@ class EnquiryViewTestCase(test_utils.BaseEnquiryTestCase):
         response = self.client.get(reverse("enquiry-list"))
         assert response.get("Content-Disposition") is None
 
+    def test_enquiry_csv_response_fields(self):
+        """
+        Asserts that response to a ``format=csv`` request returns the expected enquiry fields.
+        """
+        response = self.client.get(reverse("enquiry-list"), dict(format="csv"))
+        assert response.content.decode().strip() == ','.join(
+            settings.EXPORT_OUTPUT_FILE_CSV_HEADERS
+        )
+
     @pytest.mark.skip(reason="@TODO need to investigate why the Owner model cannot be serialized")
     def test_enquiry_list_filtered_unassigned(self):
         """Test retrieving enquiry list and ensure we get expected count"""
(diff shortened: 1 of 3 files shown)


app/enquiries/views.py
```diff
@@ -251,6 +251,13 @@ class Meta:
         }
 
 
+class EnquiryListCSVRenderer (CSVRenderer):
```
ambrozic, line 254: there is a space after class name

app/enquiries/tests/test_views.py
```diff
@@ -562,6 +562,15 @@ def test_enquiry_csv_format_content_disposition(self):
         response = self.client.get(reverse("enquiry-list"))
         assert response.get("Content-Disposition") is None
 
+    def test_enquiry_csv_response_fields(self):
+        """
+        Asserts that response to a ``format=csv`` request returns the expected enquiry fields.
+        """
+        response = self.client.get(reverse("enquiry-list"), dict(format="csv"))
+        assert response.content.decode('UTF-8').strip() == ','.join(
```
ambrozic, line 570: `utf-8` is default value for encoding. no need to specify it.

ambrozic (changes requested):

rafenden (approved):

rafenden (approved):
Pending Gregor's comments.

app/enquiries/views.py
rachaelcodes: I might also have a look into why Flake8 didn't pick that up.

ambrozic: let me know what was wrong. tried it in datahub api repo and flake8 also ignores this issue.

ambrozic (approved):

PR #178, opened 2020-08-24
Add command to mark stale enquiries as 'Non-responsive'

rachaelcodes:
## Description of change
Adds a command that checks whether an enquiry has not been updated in six weeks and has a status of 'Awaiting response'. In this case, the enquiry is marked as 'Non responsive'.

Also adds a celery task to run this command daily.

## Test instructions
If running locally, check the numbers of enquiries with 'Awaiting response' and 'Non responsive' added using the filters.
Run the command with `python manage.py mark_non_responsive_enquiries`, and note the number of enquiries changed.
Check the new number of enquiries with 'Awaiting response' and 'Non responsive'.

### DIFF
--- a/app/enquiries/celery.py
+++ b/app/enquiries/celery.py
@@ -15,6 +15,8 @@ app.autodiscover_tasks()
 
 DH_METADATA_FETCH_INTERVAL_HOURS = settings.DATA_HUB_METADATA_FETCH_INTERVAL_HOURS
 AS_ENQUIRIES_POLL_INTERVAL_MINS = settings.ACTIVITY_STREAM_ENQUIRY_POLL_INTERVAL_MINS
+ENQUIRY_STATUS_UPDATE_INTERVAL_DAYS = settings.ENQUIRY_STATUS_UPDATE_INTERVAL_DAYS
+ENQUIRY_STATUS_SHOULD_UPDATE = settings.ENQUIRY_STATUS_SHOULD_UPDATE
 
 app.conf.beat_schedule = {
     "refresh-datahub-metadata": {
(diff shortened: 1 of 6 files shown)


app/enquiries/tasks.py
```diff
@@ -27,3 +28,11 @@ def fetch_new_enquiries():
 
     fetch_and_process_enquiries()
     logging.info(f"New enquiries last retrieved at {datetime.now()}")
+
+
+@app.task(name="update_enquiry_stage")
+def update_enquiry_stage():
```
mforner13, line 34: I think the task name could be a bit more specific so it's easy to recognise its purpose in the logs - e.g. `update_stage_stale_enquiries`

sample_env
```diff
@@ -28,6 +29,7 @@ DATA_HUB_HAWK_KEY=hawk-key
 
 # Celery and Redis
 DATA_HUB_METADATA_FETCH_INTERVAL=4      # in hours
+ENQUIRY_STATUS_UPDATE_INTERVAL_DAYS=1
 REDIS_BASE_URL=redis://localhost:6379
```
mforner13, line 33: Could update the base url in a separate commit in this PR? 😃

rachaelcodes: haha great minds - I'd put it in and not pushed the changes :D

mforner13 (approved):
Looks good! 👍  Just one small thing - the commit message could give more information about what the changes are doing like in the PR title

app/enquiries/celery.py
```diff
@@ -15,6 +15,7 @@
 
 DH_METADATA_FETCH_INTERVAL_HOURS = settings.DATA_HUB_METADATA_FETCH_INTERVAL_HOURS
 AS_ENQUIRIES_POLL_INTERVAL_MINS = settings.ACTIVITY_STREAM_ENQUIRY_POLL_INTERVAL_MINS
+ENQUIRY_STATUS_UPDATE_INTERVAL_DAYS = settings.ENQUIRY_STATUS_UPDATE_INTERVAL_DAYS
```
ambrozic, line 18: This variable suggests to define and interval in days but is later used to schedule a task on a day in month. Would variable name like `ENQUIRY_STATUS_UPDATE_DAY_IN_MONTH` be more clear? Or maybe even `ENQUIRY_STATUS_UPDATE_TASK_SCHEDULE='*/15'`?

ambrozic (changes requested):

app/enquiries/celery.py
rachaelcodes: It's used with '*/' for every X days per 'day of the month', but if it's not clear, I'll update it to `ENQUIRY_STATUS_UPDATE_DAILY_SCHEDULE='*/1'` and then also update the other interval names to keep them consistent.

ambrozic: is the task suppose to be running every X days or once a month on Xth day in month?

rachaelcodes: Every X days - but day={frequency} doesn't seem to be an option.

ambrozic (approved):

PR #181, opened 2020-08-25
Make owner filter width match other filters

mforner13:
## Description of change

This PR makes the width of the 'Owner' filter on the homepage the same width as the other filters by overriding the `max-width: 100%` on `govuk-select`. 

I have added a new `owner-filter` class instead of adding the width directly to `govuk-select` because that is used by selects in the edit form which do not need a width of `100%`.

## Test instructions

- Run `npm run sass` to update the css
- Hard refresh the page
- Check that the owner filter is the same width as the others

**Before**
Screenshot 2020-08-25 at 12 17 58

**After**
Screenshot 2020-08-25 at 12 18 25

### DIFF
--- a/sass/_enquiry_list.scss
+++ b/sass/_enquiry_list.scss
@@ -65,6 +65,10 @@ $govuk-assets-path: '/static/assets/';
     padding: 0;
 }
 
+.filters__select {
+    width: 100%;
+}
+
 .import-enquiries-button {
     float: right;
     margin-bottom: 0;
@@ -81,4 +85,9 @@ $govuk-assets-path: '/static/assets/';
 
 .download-button-container {
     margin-left: auto;
+}
+
+// Make select element for owner filter same width as other filters
+.owner-filter {
+    width: 100%;
 }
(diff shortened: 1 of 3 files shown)


rachaelcodes (approved):

sass/_enquiry_list.scss
```diff
@@ -81,4 +81,9 @@ $govuk-assets-path: '/static/assets/';
 
 .download-button-container {
     margin-left: auto;
-}
\ No newline at end of file
+}
+
+// Make select element for owner filter same width as other filters
+.owner-filter {
```
peterhudec, line 87: I would select `.govuk-select` here, rather than having a rule for a specific field. How do the other `.govuk-selec`s look like? Or is this the only one?

peterhudec (approved):
Looks, good, but I would consider making the fix more general.

sass/_enquiry_list.scss
mforner13: After discussing this I've updated to a more generic `.filters__select` to be more future proof.

PR #183, opened 2020-08-25
Fix padding for summary list key items

mforner13:
## Description of change

When an item in the enquiry details summary list has no value, the key text doesn't have any top padding, this PR removes the override padding added previously which had caused this problem, and adds just padding to the left of the text.

## Test instructions

- Remove the 'website' value from one of the test fixtures
- Run `npm run sass`
- Hard refresh the page
- Go to an enquiry's details page and check that the website label has padding above it

**Before**
Screenshot 2020-08-25 at 15 33 46

**After**
Screenshot 2020-08-25 at 15 34 40

### DIFF
--- a/sass/_enquiry_detail.scss
+++ b/sass/_enquiry_detail.scss
@@ -19,6 +19,6 @@ $govuk-assets-path: '/static/assets/';
 .govuk-summary-list {
     border-top: 2px solid govuk-colour("mid-grey");
     &__key {
-        padding-bottom: 9.5px;
+        padding-left: govuk-spacing(2);
     }
 }
--- a/sass/_settings.scss
+++ b/sass/_settings.scss
@@ -38,7 +38,3 @@ ol {
 .govuk-label {
     font-weight: bold;
 }
-
-.govuk-summary-list__key {
-    padding: 0 0.5em;
-}


peterhudec (changes requested):
The actual fix should be to remove [these three lines](https://github.com/uktrade/enquiry-mgmt-tool/blob/master/sass/_settings.scss#L42-L44). What they did was confuse the horizontal and vertical padding, but I don't see a reason why they even wanted to do it, because the component was already styled properly, so the best fix is to remove the cause, which is the managed team's tampering with the govuk styles.
```scss
.govuk-summary-list__key {
    // Should have been: 0.5em 0;
    padding: 0 0.5em;
}
```

mforner13:
> The actual fix should be to remove [these three lines](https://github.com/uktrade/enquiry-mgmt-tool/blob/master/sass/_settings.scss#L42-L44). What they did was confuse the horizontal and vertical padding, but I don't see a reason why they even wanted to do it, because the component was already styled properly, so the best fix is to remove the cause, which is the managed team's tampering with the govuk styles.
> 
> ```css-scss
> .govuk-summary-list__key {
>     // Should have been: 0.5em 0;
>     padding: 0 0.5em;
> }
> ```

Thanks for finding that @peterhudec !

I have checked with Aleesha, and she has recommended we keep the alternating colours in the summary table, and add the padding on the left of the text. I've removed the code you mentioned above and made this update

peterhudec (approved):
👍

PR #194, opened 2020-09-11
Change the way DataHub metadata are fetched, cached and error handled

peterhudec:
## Description of change

This PR fixes the problematic fetching, caching and error handling of DataHub metadata and greatly simplifies it. Before this PR, the 10 metadata requests were all done at once (not in parallel though) in the [`dh_fetch_metadata`](https://github.com/uktrade/enquiry-mgmt-tool/blob/master/app/enquiries/common/datahub_utils.py#L134) and then all of them combined into a single `dict` which was subsequently cached from within the `dh_load_metadata` function. The failed requests were handled so, that a name of the failed metadata was added to a `list` in the `'failed'` key of the resulting `dict`, which broke the code which was using the `dh_fetch_metadata` for the duration of the expiration time of the cached `dict`. The result was that the real error for the _submit to DataHub_ functionality was not working was never reported.

This PR replaces the `dh_fetch_metadata` function with a `fetch_metadata` function which only fetches a single metadata endpoint at a time. The function is then called on demand where the given metadata is needed. If the underlying reques fails, the function raises an exception with [`response.raise_for_status()`](https://requests.readthedocs.io/en/master/api/#requests.Response.raise_for_status).

## Test instructions

1. Open an enquiry which can be sumbitted to DataHub
2. Click the _Submit to DataHub_ button
3. Correct all the eventual data related errors
4. The enquiry should finally be succesfully submitted to DataHub
5. Open your `.env` file and either change the `DATA_HUB_METADATA_URL` so that it's invalid, or alternatively tweak one of the `DATA_HUB_HAWK_ID` or `DATA_HUB_HAWK_KEY` values to make the metadata requests fail.
6. Restart the application so that the changes take effect
7. Just to be sure, clear the Django cache with `docker-compose exec  app python manage.py clear_cache`
8. Repeat steps 1 to 4
9. The application should render a `404` or `401` error, depending on the way you made the metadata requests fail.
10. The error related to the failed request should appear in the log.


ambrozic:
how many instances of API are we usually running in our paas?

app/enquiries/common/datahub_utils.py
```diff
@@ -95,74 +81,36 @@ def dh_request(
...
-            cached_metadata = _dh_fetch_metadata()
-            cache.set(cache_key, cached_metadata, timeout=expiry_secs)
-            return cached_metadata
-
-        logging.info(f"Metadata valid in cache (expiry_secs={expiry_secs})")
-        return cached_metadata
-    except Exception as e:
-        logging.error(f"Error fetching metadata, {str(e)} ...")
-        raise e
+    :returns: :class:`requests.Response`
```
ambrozic, line 95: Does not this function return a list of dictionaries?

ambrozic (changes requested):

app/enquiries/common/datahub_utils.py
peterhudec: Good point, I'll fix the doc.

ambrozic (approved):

peterhudec:
> how many instances of API are we usually running in our paas?

No idea

app/enquiries/common/datahub_utils.py
```diff
@@ -460,18 +408,16 @@ def dh_enquiry_readiness(request, access_token, enquiry):
     return response
 
 
+# TODO: Get rid of this function
```
elcct, line 411: Would be good to write why. May not be obvious later on.

app/enquiries/common/requests.py
```diff
@@ -0,0 +1,9 @@
+import requests
+from cachecontrol import CacheControl
+from cachecontrol_django import DjangoCache
+
+
+cached_requests = CacheControl(
```
elcct, line 6: I wonder if we need to use an external library for that. What is the value?

mforner13 (approved):
Looks good! If possible it would be great to update the sample env with updated metadata endpoints too :) 

`# Data Hub variables when running Data Hub API locally on port 8000
DATA_HUB_METADATA_URL=http://docker.for.mac.localhost:8000/v4/metadsuf
DATA_HUB_COMPANY_SEARCH_URL=http://docker.for.mac.localhost:8000/v4/search/company
DATA_HUB_CONTACT_SEARCH_URL=http://docker.for.mac.localhost:8000/v3/search/contact
DATA_HUB_CONTACT_CREATE_URL=http://docker.for.mac.localhost:8000/v3/contact
DATA_HUB_INVESTMENT_CREATE_URL=http://docker.for.mac.localhost:8000/v3/investment
DATA_HUB_ADVISER_SEARCH_URL=http://docker.for.mac.localhost:8000/adviser/
DATA_HUB_WHOAMI_URL=http://docker.for.mac.localhost:8000/whoami/`

app/enquiries/common/datahub_utils.py
peterhudec: I'll remove the todo for now.

app/enquiries/common/datahub_utils.py
```diff
@@ -490,53 +436,66 @@ def prepare_dh_payload(
...
-    payload["client_relationship_manager"] = client_relationship_manager_id
-    payload["sector"] = map_to_datahub_id(
-        enquiry.get_primary_sector_display(), dh_metadata, "sector"
+
+    payload = dict(
+        name=enquiry.company_name,
+        investor_company=company_id,
+        description=enquiry.project_description,
+        anonymous_description=enquiry.anonymised_project_description,
+        # FIXME: Can't this just be none?
```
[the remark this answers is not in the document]
peterhudec: It was referring to something I have already fixed, but forgot to remove this comment.

PR #195, opened 2020-09-14
Ensure python buildpack version is pinned

PippoRaimondi:
## Description of change

Pin python buildpack as the newest version removed the python version our application is running against.

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,5 +1,5 @@
 ---
 applications:
 - buildpacks:
-    - python_buildpack
+    - https://github.com/cloudfoundry/python-buildpack.git#v1.7.15
     - nodejs_buildpack


ambrozic (approved):

mforner13 (commented):
@PippoRaimondi Does this mean we would no longer need to specify the python version in `runtime.txt` ?

PippoRaimondi:
> @PippoRaimondi Does this mean we would no longer need to specify the python version in `runtime.txt` ?

@mforner13 Nope, the pinned buildpack version means x versions of python are available to be chosen to spin up an environment... basically the newest buildpack version chose to remove the python version we were using...

mforner13 (approved):

PR #199, opened 2020-10-06
Increase default gunicorn timeout to 5 minutes

peterhudec:
## Description of change

Increase the default 30s gUnicorn timeout to 5 minutes as a temporary fix to the issues with the _Download all_ functionality.

### DIFF
--- a/app/gunicorn.py
+++ b/app/gunicorn.py
@@ -4,6 +4,7 @@ import os
 # Access log settings
 
 accesslog = os.environ.get("GUNICORN_ACCESSLOG", "-")
+timeout = 60 * 5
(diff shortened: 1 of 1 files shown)


ambrozic (approved):

ambrozic:
timeout is not set anywhere else ie. process start command line arguments in infrastructure etc?

rachaelcodes (approved):

peterhudec:
> timeout is not set anywhere else ie. process start command line arguments in infrastructure etc?

@ambrozic Good point, not in this codebase for sure, but maybe it is in Cloud Foundry. I will check it.

PR #200, opened 2020-10-06
Post release update for Adobe campaign

harel:
## Description of change

- handle invalid emails (avoid retry), 
- remove redundant env vars, 
- fix activity stream call (access credentials)


mforner13 (approved):

app/enquiries/common/email_campaign_utils.py
```diff
@@ -173,6 +177,16 @@ def process_enquiry(enquiry):
         )
     except AdobeCampaignRequestException as exc:
         logger.exception('Error creating staging profiles in Adobe: %s', str(exc))
+        # if the failure is known and there is no need to retry, log it.
+        if any([code in exc.message for code in TERMINAL_ERROR_CODES]):
```
ambrozic, line 181: no need for converting into `list` object. `any(a is b for a in iter)` works with any type of iterator just as well. (and most of similar core functions)

ambrozic (approved):

PR #203, opened 2020-10-07
adobe record update fix

harel:
send a full enquiry on each update to adobe to prevent adobe blanking fields out


mforner13 (approved):
LGTM 👍  Also looks a bit cleaner now with just using `data`.

peterhudec (approved):

PR #204, opened 2020-10-08
Fixes related to handling of second qualification forms

harel:
## Description of change

use an isoformated date for as search. 
check if emt was provided in second qual form. 
use just date when filtering for second qual to catch discrepancies between log date and 2nd qual submission

### DIFF
--- a/app/enquiries/common/email_campaign_utils.py
+++ b/app/enquiries/common/email_campaign_utils.py
@@ -100,11 +100,14 @@ def process_second_qualifications():
     )
     for submission in submissions:
         data = submission["_source"]["object"][settings.ACTIVITY_STREAM_ENQUIRY_DATA_OBJ]["data"]
-        process_enquiry_update(
-            emt_id=data.get('emt_id'),
-            phone=data.get('phone_number'),
-            consent=data.get('arrange_callback') == 'yes'
-        )
+        if data.get('emt_id'):
+            process_enquiry_update(
+                emt_id=data.get('emt_id'),
+                phone=data.get('phone_number'),
+                consent=data.get('arrange_callback') == 'yes'
+            )
+        else:
+            logger.warning("Emt_id not provided for second qualification form")
     # kick off the workflow to process the updates
     start_staging_workflow()
 
@@ -171,10 +174,6 @@ def process_enquiry_update(emt_id, phone=None, consent=None):
     record referencing an emt_id with an updated phone and consent,
     and new stage, to update the relevant profile.
     """
-    log = None
-    if not emt_id:
-        logger.warning("Emt_id not provided for second qualification form")
-        return log
     data = {
         'phone': phone,
         'phoneConsent': consent,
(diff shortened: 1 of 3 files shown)


app/enquiries/common/email_campaign_utils.py
```diff
@@ -179,6 +184,13 @@ def process_enquiry_update(emt_id, phone=None, consent=None):
     client = AdobeClient()
     try:
         enquiry = Enquiry.objects.get(id=emt_id)
+        # check if this has already been processed
+        has_processed = EnquiryActionLog.objects.filter(
+            enquiry=enquiry,
+            action=ref_data.EnquiryAction.SECOND_QUALIFICATION_FORM
+        ).exists()
+        if has_processed:
+            return log
```
mforner13, line 193: What's the purpose of returning `None` here and on line 177? If the return value doesn't matter can we just `return`?

harel: It's just a habit. In a happy path it returns the log object , which is useful if you're testing or debugging. So as a habit if i return something, i try to return None explicitly if there is a None value to return . 
I can change to an empty return if that works better for you though. i don't mind

peterhudec: The return value doesn't seem to be used apart from a test, where `None` is asserted, so the `log` variable can be removed.

app/enquiries/tests/test_adobe_integration.py
```diff
@@ -182,3 +182,8 @@ def test_log_action(self):
 
         last_action_date = EnquiryActionLog.get_last_action_date(action)
         self.assertEqual(last_action_date.enquiry.id, self.enquiry.id)
+
+    @freeze_time()
+    def test_process_enquiry_update_bad_id(self):
+        log = campaign.process_enquiry_update(None)
+        assert log is None
```
peterhudec, line 189: This kind of tests are of little value as it only checks one specific scenario. I would remove it.

app/enquiries/common/email_campaign_utils.py
```diff
@@ -170,6 +172,9 @@ def process_enquiry_update(emt_id, phone=None, consent=None):
     and new stage, to update the relevant profile.
     """
     log = None
+    if not emt_id:
+        logger.warning("Emt_id not provided for second qualification form")
+        return log
```
peterhudec, line 177: This check should be done outside of the function and the function should only be called if `emt_id` is not `None`. The `emt_id` is after all defined as a _required_ parameter.

peterhudec (changes requested):
Looks good, just a couple of things.

app/enquiries/tests/test_adobe_integration.py
harel: i only tried to get it in because coverage was complaining.

mforner13: As codecov passing is not mandatory for the PR to be merged, I'd say best to remove this if it's only to placate the check, so as not to give false reassurance?

mforner13 (approved):

peterhudec (approved):

PR #206, opened 2020-10-14
Enquiry ID filter

harel:
Add an enquiry id filter. This helps users locate an enquiry based on
the id and is useful for locating submissions of second qualification
forms.

## Description of change

Adds a new filter search field, and subsequent filter definition

## Test instructions

In the enquiries screen a new field should show at the bottom of the filter list.

### DIFF
--- a/app/enquiries/templates/partials/enquiry_filters.html
+++ b/app/enquiries/templates/partials/enquiry_filters.html
@@ -62,6 +62,12 @@
     </label>
     <input class="govuk-input" id="project_code__icontains" name="project_code__icontains" type="search" value="{{query_params.project_code__icontains}}">
   </div>
+  <div class="govuk-form-group">
+    <label class="govuk-label" for="project_code__icontains">
+      Enquiry ID
+    </label>
+    <input class="govuk-input" id="emt_id" name="id" type="search" value="{{query_params.id}}">
+  </div>
   <button id="btn_submit" class="govuk-button govuk-!-margin-right-1" data-module="govuk-button" accesskey="s">
     Apply filters
   </button>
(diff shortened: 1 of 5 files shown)


ambrozic (approved):

app/enquiries/templates/partials/enquiry_filters.html
```diff
@@ -62,6 +62,12 @@
     </label>
     <input class="govuk-input" id="project_code__icontains" name="project_code__icontains" type="search" value="{{query_params.project_code__icontains}}">
   </div>
+  <div class="govuk-form-group">
+    <label class="govuk-label" for="project_code__icontains">
```
mforner13, line 66: The `for` needs updating to match `emt_id`

app/enquiries/views.py
```diff
@@ -207,6 +207,7 @@ class EnquiryFilter(filters.FilterSet):
     """
     Enquiry search filters
     """
+    id = filters.NumberFilter(field_name="id", method="filter_id")
```
mforner13, line 210: Could specify `enquiry__id` just to make it super clear? And given we've got `owner__id` too.

mforner13 (changes requested):
A couple of suggestions, but mainly there needs to be tests added for the new filter. 

There's functional tests for the other filters in `filter.test.js`and unit in `test_filters.py`.

app/enquiries/views.py
harel: This would break the filter i think.... it's based on the field name and it's path in the objects.

mforner13 (approved):

PR #210, opened 2020-10-28
Rest API endpoints with hawk authentication

ambrozic:
## Description of change

- added Rest API with hawk authentication
- added rest api endpoint for enquiries data

## Test instructions

### DIFF
--- a/app/enquiries/tests/test_auth.py
+++ b/app/enquiries/tests/test_auth.py
@@ -1,0 +1,462 @@
+import datetime
+from collections.abc import Mapping, Sequence
+
+import mohawk
+import pytest
+from django.urls import reverse, path
+from freezegun import freeze_time
+from rest_framework import status
+from rest_framework.response import Response
+from rest_framework.test import APIRequestFactory
+from rest_framework.views import APIView
+
+from app.enquiries import auth
+from app.enquiries.tests.factories import OwnerFactory
+
+HAWK_CREDENTIALS = {
+    "test-id-without-scope": {
+        "key": "test-key-without-scope",
+        "scopes": (),
+    },
+    "test-id-with-scope": {
+        "key": "test-key-with-scope",
+        "scopes": ("enquiries",),
+    },
+    "test-id-with-multiple-scopes": {
+        "key": "test-key-with-multiple-scopes",
+        "scopes": ("enquiries", "views"),
+    },
+}
+
+
+class HawkViewWithoutScope(auth.HawkResponseSigningMixin, APIView):
+    """View using Hawk authentication."""
+
+    authentication_classes = (auth.HawkAuthentication,)
+    permission_classes = ()
+
+    def get(self, request):
+        """Simple test view with fixed response."""
+        return Response({"content": "hawk-test-view-without-scope"})
+
+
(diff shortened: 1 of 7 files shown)


app/enquiries/tests/test_auth.py
```diff
@@ -0,0 +1,461 @@
...
+        response = api_client.get(
+            _url(),
+            content_type="",
+            HTTP_AUTHORIZATION=sender.request_header,
+        )
+
+        assert response.status_code == status.HTTP_200_OK
+        assert response.json() == {"content": "hawk-test-view-without-scope"}
+
+    def test_empty_object_returned_with_authentication(self, api_client, settings):
```
mforner13, line 300: In these two tests it looks like you're making sure an object which is **not** empty is returned, so the names seem a bit misleading. Apologies if I've misunderstood.

app/enquiries/tests/test_views.py
```diff
@@ -717,3 +713,129 @@ def test_import_template(self):
...
+@pytest.mark.django_db
+class TestAPIEnquiries:
+    def test_enquiries_list_unauthorized(self, api_client):
+        """Should return 401"""
+        url = reverse("api-v1-enquiries")
+        response = api_client.get(url)
+        assert response.status_code == status.HTTP_401_UNAUTHORIZED
+
+    def test_enquiries_list_forbidden(self, api_client, settings):
+        """Should return 200"""
```
mforner13, line 749: Looks like it should return a 403?

mforner13 (commented):
A few comments, + adding adding the hawk credential env vars to `sample.env` :)

app/enquiries/views.py
```diff
@@ -93,6 +97,13 @@ def get_enquiry_field(name):
     return {"name": name, "choices": filter_config[name].choices}
 
 
+class Pagination(PageNumberPagination):
+    """
+    Control API views pagination.
+    """
+    page_size_query_param = settings.ENQUIRIES_PAGE_SIZE_PARAM
```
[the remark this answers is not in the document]
ambrozic: i want to control page size per request ie `enquiries?page_size=123` and i could not find a better way to do it. ideally this should be a setting in `settings.REST_FRAMEWORK`

app/settings/common.py
```diff
@@ -227,6 +224,7 @@
 # App specific settings
 CHAR_FIELD_MAX_LENGTH = 255
 ENQUIRIES_PER_PAGE = env.int('ENQUIRIES_PER_PAGE', default=10)
+ENQUIRIES_PAGE_SIZE_PARAM = env.int('ENQUIRIES_PAGE_SIZE_PARAM', default='page_size')
```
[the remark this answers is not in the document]
ambrozic: same as above, i wanted to define a query parameter name that controls page size on request.

app/enquiries/tests/test_auth.py
ambrozic: thank you. renamed. was copy pasting tests from DH API :)

mforner13 (approved):
👍

elcct (approved):

PR #211, opened 2020-11-27
Bugfix: prevent duplicate contact creation

rachaelcodes:
## Description of change
Fixes a bug whereby if company contacts known to Data Hub made enquiries that were fed through the EMT, a duplicate contact would be created.

Now when preparing to submit to Data Hub, there is a check to see if the enquirer's first name, last name and email is an exact match to any of the existing contacts in Data Hub (business logic here agreed by Nitesh). If there is an exact match, this contact's id is used in submitting the enquiry to DH, otherwise a new company contact is created first. 

If there are no existing contacts for the company, the new contact created here will be labeled as a 'Primary' contact.

## Test instructions
If running locally using the default api fixtures: 

- create a new enquiry
- link this enquiry to 'Venus Ltd' in your local DH instance
- make sure the enquirer's name is: `Johnny Cakeman` with the email address: `johnny@cakeman.com`
- add the enquiry to DataHub (locally)
- check that the enquiry has been added to DH
- check that a duplicate Johnny Cakeman has not been added to the database

### DIFF
--- a/app/enquiries/common/datahub_utils.py
+++ b/app/enquiries/common/datahub_utils.py
@@ -195,18 +195,15 @@ def dh_company_search(request, access_token, company_name):
     return companies, None
 
 
-def dh_contact_search(request, access_token, contact_name, company_id):
+def dh_get_company_contact_list(request, access_token, company_id):
     """
-    Performs a `contact name` search using |data-hub-api|_.
+    Performs a contact search for a specific company using |data-hub-api|_.
 
     :param request:
     :type request: django.http.HttpRequest
 
     :param access_token: The user's |oauth|_ access token
     :type access_token: str
-
-    :param contact_name: Contact name to search for
-    :type contact_name: str
 
     :param company_id: Company ID
     :type company_id: str
@@ -216,7 +213,7 @@ def dh_contact_search(request, access_token, contact_name, company_id):
     """
     contacts = []
     url = settings.DATA_HUB_CONTACT_SEARCH_URL
-    payload = {"name": contact_name, "company": [company_id]}
+    payload = {"company": [company_id]}
 
     response = dh_request(request, access_token, "POST", url, payload)
 
@@ -238,6 +235,36 @@ def dh_contact_search(request, access_token, contact_name, company_id):
     return contacts, None
 
 
+def dh_get_matching_company_contact(first_name, last_name, email, company_contacts):
+    """
+    Performs check identifying whether an enquirer exists in their company's contact
+    list in |data-hub-api|_.
+
+    :param first_name:
+    :type first_name: str
+
+    :param last_name:
+    :type last_name: str
+
+    :param email:
+    :type email: str
+
+    :param company_contacts:
+    :type company_contacts: list
+
+    :returns: The first matching contact if any exist
+    :rtype: dict or None
+    """
+    return next((
+            company_contact for company_contact in company_contacts
+            if company_contact["first_name"].lower() == first_name.lower()
+            and company_contact["last_name"].lower() == last_name.lower()
+            and company_contact["email"].lower() == email.lower()
+        ),
+        None
+    )
+
+
 def dh_contact_create(request, access_token, enquirer, company_id, primary=False):
     """
     Create a |data-hub|_ `contact` and associate it with the given `company`.
@@ -248,8 +275,8 @@ def dh_contact_create(request, access_token, enquirer, company_id, primary=False
     :param access_token: The user's |oauth|_ access token
     :type access_token: str
 
-    :param contact_name: Contact name to search for
-    :type contact_name: str
+    :param enquirer: Contact to add
+    :type enquirer: app.enquiries.models.Enquirer
 
     :param company_id: Company ID
     :type company_id: str
@@ -261,7 +288,6 @@ def dh_contact_create(request, access_token, enquirer, company_id, primary=False
     :rtype: tuple
     """
     url = settings.DATA_HUB_CONTACT_CREATE_URL
-    enquirer = enquirer.enquirer
     payload = {
         "first_name": enquirer.first_name,
         "last_name": enquirer.last_name,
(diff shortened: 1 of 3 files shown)


app/enquiries/common/datahub_utils.py
```diff
@@ -195,19 +195,16 @@ def dh_company_search(request, access_token, company_name):
     return companies, None
 
 
-def dh_contact_search(request, access_token, contact_name, company_id):
```
rachaelcodes, line 198: This wasn't doing what was described: there's a fuzzy match on the name in the DH search, so it could falsely identify duplicates.

The output of this function was also being used to identify not whether there were matching contacts in the company, but any contacts at all - which is why I've removed the contact name parameter.

app/enquiries/common/datahub_utils.py
```diff
@@ -216,7 +213,7 @@ def dh_contact_search(request, access_token, contact_name, company_id):
     """
     contacts = []
     url = settings.DATA_HUB_CONTACT_SEARCH_URL
-    payload = {"name": contact_name, "company": [company_id]}
```
rachaelcodes, line 219: see comment above for why this was removed

app/enquiries/common/datahub_utils.py
```diff
@@ -261,7 +288,6 @@ def dh_contact_create(request, access_token, enquirer, company_id, primary=False
     :rtype: tuple
     """
     url = settings.DATA_HUB_CONTACT_CREATE_URL
-    enquirer = enquirer.enquirer
```
rachaelcodes, line 264: It seemed clearer to just pass the enquiry.enquirer object in directly

app/enquiries/common/datahub_utils.py
```diff
@@ -541,27 +625,11 @@ def dh_investment_create(request, enquiry):
         return response
 
     company_id = enquiry.dh_company_id
-    # Create a contact for this company
-    # If a contact already exists then make the new contact as secondary
-    full_name = f"{enquiry.enquirer.first_name} {enquiry.enquirer.last_name}"
-    existing_contacts, error = dh_contact_search(request, access_token, full_name, company_id)
```
rachaelcodes, line 547: I'm not sure it made sense to pass in an enquirer name to a search that was used to see whether there were any contacts in that company.

app/enquiries/tests/test_dh_integration.py
```diff
@@ -59,7 +59,7 @@ def contact_search_response():
                 }
             ]
         },
-        "error": {"name": ["This field may not be blank."]},
```
rachaelcodes, line 62: Updated to reflect the actual api error messages

app/enquiries/tests/test_dh_integration.py
```diff
@@ -163,32 +163,21 @@ def test_company_search_error(self):
...
+            response, error = dh_get_company_contact_list(
+                "mock_request", "access_token", "company_id"
             )
             self.assertIsNone(error)
             self.assertEqual(len(response), 1)
             self.assertEqual(response[0]["datahub_id"], expected[0]["id"])
             self.assertEqual(response[0]["first_name"], expected[0]["first_name"])
             self.assertEqual(response[0]["last_name"], expected[0]["last_name"])
 
-    def test_contact_search_error(self):
```
rachaelcodes, line 181: Removed as this is not how the api seems to behave

app/enquiries/common/datahub_utils.py
```diff
@@ -281,6 +307,64 @@ def dh_contact_create(request, access_token, enquirer, company_id, primary=False
     return response.json(), None
 
 
+def prepare_dh_contact(request, access_token, enquirer, company_id):
```
ambrozic, line 310: could this be renamed to `dh_contact_prepare` for the sake of naming consistency in this module?

app/enquiries/common/datahub_utils.py
```diff
@@ -281,6 +307,64 @@ def dh_contact_create(request, access_token, enquirer, company_id, primary=False
...
+    :type request: django.http.HttpRequest
+
+    :param access_token: The user's |oauth|_ access token
+    :type access_token: str
+
+    :param enquirer:
+    :type enquirer: app.enquiries.models.Enquirer
+
+    :param company_id:
+    :type company_id: uuid
```
ambrozic, line 325: is `company_id` `uuid` or `str`? in code and in tests it is used and passed around as `str`

app/enquiries/common/datahub_utils.py
```diff
@@ -281,6 +307,64 @@ def dh_contact_create(request, access_token, enquirer, company_id, primary=False
...
+    :returns: A ``(uuid, None)`` or ``(None, error_json_dict)`` tuple
+    :rtype: tuple
+    """
+
+    # Check whether enquiry company has existing contacts in DH
+    existing_contacts, error = dh_get_company_contact_list(
+        request, access_token, company_id
+    )
+    if error:
+        return None, {"contact_search": f"Error while checking company contacts, {str(error)}"}
```
ambrozic, line 336: does `error` have to be cast to `str`?

ambrozic (changes requested):

ambrozic:
is it possible there will be issues, at least in emails, with case sensitivity?

app/enquiries/common/datahub_utils.py
rachaelcodes: Yes, there's a `prepare_dh_payload` elsewhere in the file which this was copying, but what I'll do is bring them both round to this style.

app/enquiries/common/datahub_utils.py
rachaelcodes: This is the pattern through the rest of the 'submit to DataHub' functions - I think the idea is that the error messages are passed as strings to view to populate the error information box for users.

rachaelcodes:
> is it possible there will be issues, at least in emails, with case sensitivity?

Good shout. I'll cancel out the case on the names too, as some are entered in all-caps, and sometimes people are lazy with capital letters.

app/enquiries/common/datahub_utils.py
ambrozic: in the similar code below errors are not casted to string. error object seem to be dict or exception and when string-formatted, this objects are converted to string by default without issues.

ambrozic (approved):

app/enquiries/common/datahub_utils.py
```diff
@@ -238,6 +235,36 @@ def dh_contact_search(request, access_token, contact_name, company_id):
...
+    :type company_contacts: list
+
+    :returns: A subset of fields for each company found
+    :rtype: dict or None
+    """
+    return next((
+            company_contact for company_contact in company_contacts
+            if company_contact["first_name"].lower() == first_name.lower()
+            and company_contact["last_name"].lower() == last_name.lower()
+            and company_contact["email"].lower() == email.lower()
```
mcbhenwood, line 262: I'd argue that the name checks are redundant if the email matches, but I note the agreed logic in the PR header so OK.

app/enquiries/common/datahub_utils.py
```diff
@@ -238,6 +235,36 @@ def dh_contact_search(request, access_token, contact_name, company_id):
...
+    :param last_name:
+    :type last_name: str
+
+    :param email:
+    :type email: str
+
+    :param company_contacts:
+    :type company_contacts: list
+
+    :returns: A subset of fields for each company found
```
mcbhenwood, line 255: Returns a matching contact.

mcbhenwood (approved):

PR #214, opened 2021-01-13
Consent service integration

ambrozic:
## Description of change

Change includes reading and writing consents using Consent Service. Reading consents from consent service happen in real-time on every request to view Enquiry, write includes updating consents using celery task on every Enquiry create or update. This solution currently works in parallel with existing approach until data is later synced and existing functionality removed.

[trello ticket](https://trello.com/c/RRrepX66/497-emt-consent-service-1-update-consent-service-phone-email)

## Test instructions


app/enquiries/common/consent.py
```diff
@@ -0,0 +1,70 @@
...
+        ),
+        default_timeout=(
+            settings.CONSENT_SERVICE_CONNECT_TIMEOUT,
+            settings.CONSENT_SERVICE_READ_TIMEOUT,
+        ),
+    )
+    return client.request(path=url, method=method, **kwargs)
+
+
+def check(key):
```
mforner13, line 37: Even though they are in the `consent.py` file, I think adding `consent` to these two functions would avoid any possibility of confusion :) E.g `check_consent` & `set_consent`

app/enquiries/common/client.py
```diff
@@ -0,0 +1,69 @@
...
+                method,
+                url,
+                auth=self._auth,
+                headers=headers,
+                timeout=timeout,
+                **kwargs,
+            )
+        except ConnectionError as e:
+            logger.exception(e)
+            raise BadGatewayAPIException(
```
elcct, line 63: Would be good to test behaviour when this is raised.

elcct (approved):

app/enquiries/common/consent.py
```diff
@@ -0,0 +1,70 @@
...
+
+    key = key.lower()
+    key_type = "email" if "@" in key else "phone"
+
+    data = {
+        "consents": [f"{key_type}_marketing"] if value else [],
+        key_type: key,
+        "modified_at": datetime.now().isoformat(),
+    }
+
```
mforner13, line 64: Out of curiosity, why did you chose to make it only handle an email or phone at a time? Rather than it making one request to update both fields? (if both are changing)

mforner13 (commented):
Looks good! Just two comments :)

app/enquiries/common/consent.py
ambrozic: i expected this comment from you 😄 will fix it.

app/enquiries/common/client.py
ambrozic: Copied all tests for `HawkAUth` and `APIClient` straight from `DH API`.

app/enquiries/common/consent.py
mforner13: Haha Miriam aka the naming police 🚓  👮‍♀️

mforner13 (approved):

PR #215, opened 2021-01-27
Consent service integration

ambrozic:
## Description of change

This piece of work is second part of consent service integration and is only removing consent fields from `Enquirer` model.

## Test instructions


app/enquiries/tasks.py
```diff
@@ -11,7 +11,28 @@
     process_second_qualifications,
     process_engaged_enquiries,
 )
-from app.enquiries.utils import mark_non_responsive_enquiries
+from app.enquiries.utils import str2bool, mark_non_responsive_enquiries
+
+
+def set_enquirer_consent_update(data: dict):
```
mforner13, line 17: Does this need the `@app.task(name="")` decorator to register the task? 

Also I know you knew I was gunna say this, but a description of what the task does as the others have would be useful 😛

app/enquiries/common/as_utils.py
```diff
@@ -123,11 +123,11 @@ def via_enquiry_to_enquiry_kwargs(data):
...
     enquiry["enquirer"]["email"] = email
     enquiry["enquirer"]["phone"] = data["Phone number"]
-    email_consent = strtobool(
+    email_consent = utils.strtobool(
         data.get("I would like to receive additional information by email", "False")
     )
-    phone_consent = strtobool(
-        data.get("I would like to receive additional information by telephone", "False",)
+    phone_consent = utils.strtobool(
+        data.get("I would like to receive additional information by telephone", "False")
```
mforner13, line 130: Should this be str2bool to reflect the function you've added to utils?

app/enquiries/common/as_utils.py
```diff
@@ -312,6 +312,8 @@ def fetch_and_process_enquiries():
 
     logging.info(f"Total number of enquiries retrieved: {len(enquiries)}")
 
+    from app.enquiries import tasks
```
mforner13, line 315: Could this import go at the top so they're all together? :)

app/enquiries/tasks.py
ambrozic: this function is not a task. it is a utility function inside tasks module to create tasks that and is used in various places throughout the project. ie [here](/uktrade/enquiry-mgmt-tool/pull/215/files#diff-dc32ffed8ba8cfde12b4cda9369ae0608338a2a4db4ebba34871559f053a78eaR327), [here](https://github.com/uktrade/enquiry-mgmt-tool/pull/215/files#diff-0f54edd3abaa92a34ab9e0f91b9cd75dc6b7ee25b892b4f57ee1f5466830aeabR370) and [here](https://github.com/uktrade/enquiry-mgmt-tool/pull/215/files#diff-0f54edd3abaa92a34ab9e0f91b9cd75dc6b7ee25b892b4f57ee1f5466830aeabR454). will add description, it will help clear things out :)

app/enquiries/common/as_utils.py
ambrozic: thats a good catch. thank you 😄

app/enquiries/common/as_utils.py
ambrozic: It would create cyclic import because how modules are organised. i was forced to do a local import.

mforner13 (approved):

app/enquiries/tasks.py
mforner13: The move to the utils file looks good and will avoid any confusion! 👍

elcct (approved):

PR #216, opened 2021-02-04
Fixed processing of invalid form submissions

ambrozic:
## Description of change
Invalid data from forms was causing tasks to break and enquiries were left unprocessed since October. Invalid forms are now ignored following business decision.

### DIFF
--- a/app/enquiries/tests/test_adobe_integration.py
+++ b/app/enquiries/tests/test_adobe_integration.py
@@ -1,15 +1,16 @@
 import datetime
-import app.enquiries.ref_data as ref_data
+from unittest import mock
+
+from django.conf import settings
 from django.test import TestCase
 from django.utils import timezone
-from django.conf import settings
 from faker import Faker
 from freezegun import freeze_time
-from unittest import mock
-from app.enquiries.models import Enquiry, Enquirer, EnquiryActionLog
+
+import app.enquiries.ref_data as ref_data
 from app.enquiries.common import email_campaign_utils as campaign
 from app.enquiries.common.adobe import AdobeClient
-
+from app.enquiries.models import Enquiry, Enquirer, EnquiryActionLog
 
 faker = Faker()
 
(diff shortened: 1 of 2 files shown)


app/enquiries/tests/test_adobe_integration.py
```diff
@@ -142,6 +143,39 @@ def test_process_second_qualification(self, mock_staging, mock_wf, mock_as, mock
         )
         client.start_workflow.assert_called_with(settings.ADOBE_STAGING_WORKFLOW)
 
+    @freeze_time()
+    @mock.patch('app.enquiries.common.adobe.AdobeClient.get_token')
+    @mock.patch('app.enquiries.common.as_utils.hawk_request')
+    @mock.patch('app.enquiries.common.email_campaign_utils.get_new_second_qualification_forms')
+    def test_process_second_qualification_invalid_data(self, mock_as, mock_token, mock_ecu):
```
tomkingston, line 150: Are mock_token and mock_ecu the wrong way around here?

tomkingston (commented):
Looks good, but I think the arguments passed from your mock patches are in the wrong order - let me know if I've got that wrong

app/enquiries/tests/test_adobe_integration.py
ambrozic: changed tests a little. should be ok now.

tomkingston (approved):

app/enquiries/tests/test_adobe_integration.py
```diff
@@ -142,6 +143,48 @@ def test_process_second_qualification(self, mock_staging, mock_wf, mock_as, mock
...
+    @mock.patch('app.enquiries.common.adobe.AdobeClient.get_token')
+    @mock.patch('app.enquiries.common.as_utils.hawk_request')
+    def test_process_second_qualification_invalid_data(self, mock_as, mock_token):
+        for emt_id, phone_number, arrange_callback, was_called in [
+            ('None', '0771231234', 'yes', True),
+            ('-1', '0771231234', 'yes', True),
+            (1.2, '0771231234', 'yes', True),
+            (None, '0771231234', 'yes', False),
+            ("", '0771231234', 'yes', False),
+        ]:
```
tomkingston, line 156: You could use pytest.parametrize here if you wanted: https://docs.pytest.org/en/stable/parametrize.html

ambrozic: i think parametrize does not work with `django.test.TestCase`

tomkingston: Also, given that `arrange_callback` is yes every time, and the `phone_number` is also the same in each case, perhaps they could be moved out of the fixture

app/enquiries/tests/test_adobe_integration.py
```diff
@@ -142,6 +143,48 @@ def test_process_second_qualification(self, mock_staging, mock_wf, mock_as, mock
...
+                                            'arrange_callback': arrange_callback,
+                                        }
+                                    }
+                                }
+                            }
+                        ]
+                    }
+                }
+                campaign.process_second_qualifications()
+                assert mock_update.called is was_called
```
mforner13, line 180: Could this be asserted using `assert mock_udpate.called`? It could be confusing using the `was_called` value, as that is referring to a physical phone call rather than a function call, and seems unrelated. Unless that function only ever gets called if the contact was called in real life?

ambrozic: `was_called` refers to mocked function being called or executed and `mock_udpate.called` has to be asserted as a boolean

mforner13: Understood - much clearer now using booleans in the fixture data! 👍

mforner13 (approved):

PR #431, opened 2022-12-14
Chore/dependencies 2022 12 12

elcct:
## Description of change

Dependabot updates for 12/12/2022

## Test instructions

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,7 +19,7 @@ services:
       DJANGO_SUPERUSER_EMAIL: testuser@example.com
       DJANGO_SUPERUSER_PASSWORD: testpass
       FEATURE_ENFORCE_CONSENT_SERVICE: 0
-    entrypoint: dockerize -wait tcp://db:5432 -wait tcp://redis:6379 -timeout 5m
+    entrypoint: dockerize -wait tcp://db:5432 -wait tcp://redis:6379 -wait tcp://mock-sso:8080 -timeout 5m
     command: ./start.sh
 
   celery:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,7 +33,7 @@ services:
     command: celery -A app worker -l info -Q celery -B
 
   db:
-    image: postgres
+    image: postgres:12
     environment:
       POSTGRES_DB: postgres
       POSTGRES_PASSWORD: password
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -19,7 +19,7 @@ services:
       DJANGO_SUPERUSER_EMAIL: testuser@example.com
       DJANGO_SUPERUSER_PASSWORD: testpass
       FEATURE_ENFORCE_CONSENT_SERVICE: 0
-    entrypoint: dockerize -wait tcp://db:5432 -wait tcp://redis:6379 -timeout 5m
+    entrypoint: dockerize -wait tcp://db:5432 -wait tcp://redis:6379 -wait tcp://mock-sso:8080 -timeout 5m
```
marijnkampf, line 22: Is the mock-sso supposed to be in the docker-compose.yml? Will this be used on the live server or is this used for local dev setups only?

requirements.txt
```diff
@@ -16,7 +16,7 @@ django-autocomplete-light==3.5.0
 django-cache-memoize==0.1.10
 django-environ==0.9.0
 django-extensions==3.2.1
-django-filter==2.4.0
+django-filter==22.1
```
marijnkampf, line 19: FYI for Export Wins we ran into a problem where the upgrade to 22.1 deprecated a number of fields.

See https://django-filter.readthedocs.io/en/stable/guide/migration.html#view-attributes-renamed-867

It looks like this is used on https://github.com/uktrade/enquiry-mgmt-tool/blob/master/app/enquiries/views.py#L85 (please double check that there are no other deprecated usages)

docker-compose.yml
elcct: No this is only for development and that entry ensures the `mock-sso` - which is a development dependency - is loaded before the app is. It is also being used by CircleCI to run tests. 
The actual deployment is defined here https://github.com/uktrade/enquiry-mgmt-tool/blob/master/Procfile

requirements.txt
elcct: Thank you! I think that example is a bit misleading as that function is used to build choices for enquiry stage filter and the actual filtering is defined here https://github.com/uktrade/enquiry-mgmt-tool/blob/master/app/enquiries/views.py#L225
I think it looks up to date, e.g. it uses `field_name` attributes instead of `name` for individual filters and it is assigned correctly through `filterset_class` rather than `filter_class`.
So I think everything looks okay.

marijnkampf: I only had a very quick look at it, just wanted to double check with you.

marijnkampf (approved):

PR #986, opened 2024-07-23
Rename EMT-to-Data-Hub hawk credentials for consistency

oliverjwroberts:
## Description of change

Attempting to send an enquiry from EMT to Data Hub results in a server error. Upon investigation this is due to EMT not having the correct authentication for Data Hub's Metadata endpoint, and the reason for this is two-fold:

1. A mismatch in hawk credentials
2. Data Hub not giving any scope to the existing `ENQUIRY_MGMT_HAWK_*` credentials

To fix this, in part, this PR renames these environment variables/secrets to be consistent across EMT and Data Hub. 

Please see related PR's here:

- [Data Hub Deploy PR #30](https://github.com/uktrade/data-hub-deploy/pull/30)
- [Data Hub API PR #5546](https://github.com/uktrade/data-hub-api/pull/5546)

## Test instructions

Along with the changes made in the related PRs, you should be able to successfully submit an enquiry to Data Hub.


DeanElliott96 (approved):

DeanElliott96:
There is a list of environment variables in the platform-documentation repo as well (containing the one that's now renamed). Not sure if we could also remove that since environment variables are likely to be changed?

https://platform.readme.trade.gov.uk/migrations/migration-progress/data-hub/#enquiry-management-tool

cgsunkel (approved):

oliverjwroberts:
> There is a list of environment variables in the platform-documentation repo as well (containing the one that's now renamed). Not sure if we could also remove the full list since environment variables are likely to be changed? Maybe don't want to be maintaning a double list of our environments? Might not be massively related to this PR so I've approved this one!
> 
> https://platform.readme.trade.gov.uk/migrations/migration-progress/data-hub/#enquiry-management-tool

Ahh I didn't realise these were listed in the platform docs. I agree, maintaining duplicate lists would be silly. Maybe we should raise it with the other Data Hub devs and decide where we want the list of variables to live. At the moment, it seems to vary from repo to repo, and would be good to agree on a consistent approach.

PR #1002, opened 2024-07-31
Dependencies 31-07-2024

oliverjwroberts:
## Description of change

As it says on the tin.

Dependabot also raised #995 to bump Django to 5.x, however, given that we are not using Django 5.x within Data Hub, I told it to ignore future 5.x updates (this can be reverted by reopening the PR).

## Test instructions

You should be able to run EMT as normal and all tests should pass.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,11 +12,14 @@
       "dependencies": {
         "govuk-frontend": "^5.4.0",
         "jquery": "^3.7.1",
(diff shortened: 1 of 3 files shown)


cgsunkel (approved):
Thank you for doing this

oliverjwroberts:
> Thank you for doing this

No worries at all, I'll see if I can do some others over the next few days.

